PR #14, opened 2025-01-07
predicate subscriptions working in browser. Numerous other cleanups and refactors

dnorman:
Apologies for the ginormous single commit. I tried to avoid gratuitous changes in this branch, but it took a lot to get this working.

But the good news is that subscriptions work in the browser now!

image

There are lots more cleanups and ergonomic fixes to be done, particularly around the react signals story. For the moment I've just got the `TestResultSet` wasm-bindgen wrapper hand-coded examples/wasm-bindings/src/lib.rs

```
#[wasm_bindgen]
#[derive(WasmSignal, Debug, Clone, Default)]
pub struct TestResultSet(Arc<ResultSet<SessionRecord>>);
```

That, and the subscribe need some changes to avoid boilerplate buildup when we start developing real stuff with this.


dnorman:
I approve this message ✅

PR #23, opened 2025-01-12
Misc2

dnorman:
Misc improvements to dependency management, predicates, etc


dnorman:
cc @mmacfadden @liam-germain

PR #50, opened 2025-02-07
Added a not equal comparison index.

mmacfadden:
Added a comparison index subscription for not-equal.


dnorman (approved):

dnorman:
Looks good! ComparisonIndex is a weird approach huh?

The point of this is to provide an initial implementation of a predicate-oriented inverted index.

The plan is to push this into the storage engine. Key-value stores (like sled and TiKV) tend to have an optimized range query mechanism which works for at least some of these.

The Postgres implementation will later have a table or series of tables that essentially mirror the ComparisonIndex data structure, and implement the matching logic with triggers in each table + NOTIFY

PR #54, opened 2025-02-12
PolicyAgent groundwork

dnorman:
- Add PolicyAgent trait
- Make Node generic over PolicyAgent
- Node type erasure trait for connectors
- Node type erasure for Transaction

Next steps (not in this PR):
- Implement `Context` struct for type-erased contextualized node operations (fetch/subscribe/begin) with a credential
- Model accessors for wasm. Eg: `Artist.fetch(context, "status = 'Active'")`


dnorman:
The plan is to instantiate the context struct with something like :
```
impl PolicyAgent for MyPolicyAgent {
   type ContextData = MyCredential;
   ...
}

#[wasm_bindgen]
fn get_context(data: MyCredential) -> Context {
    let node: Node<MyPolicyAgent> = existing_node(); // Generic typed node can't be exported to wasm
     node.context(data) // Type erased - wasm_bindgen is implemented 
}
```

mmacfadden:
How does this interplay with the `client`?  A lot of the current methods pass in the client, but that is really to get the node.  So just thinking through what the code would look like.  There is perhaps some room to rethink the react context provider.  I would have to have to do something like this in each method.

```ts
const client = useClient();
const my_creds = ...
const context = get_context(my_creds);

// use the context.
```
Seems like most of the time, we don't really want the client. We are just using the client to get the node.  So perhaps initializing a context somewhere else in the app (e.g. with a provider) and doing something like the below might reduce builder plate.

```ts
const context = useContext();
const artists = await Artist.fetch(context, "status = 'Active'");
```

Stuff for a future PR, likely.  just wanted to comment now.

core/src/policy.rs
```diff
@@ -0,0 +1,89 @@
...
+
+/// Applications will implement this trait to control access to resources
+/// (Entities and RPC calls) and the Node will be generic over this trait
+pub trait PolicyAgent {
+    /// The context type that will be used for all resource requests.
+    /// This will typically represent a user or service account.
+    type ContextData: ContextData;
+
+    // For checking if a context can access a collection
+    fn can_access_collection(&self, data: &Self::ContextData, collection: &CollectionId) -> AccessResult;
```
mmacfadden, line 31: Thinking through if these need to be a Result.  I suppose it could be up to an implementer to deal with this... but I suspect for some implementations might have errors. I assume an error probably equals deny.  Just curious if we handle that logic, or make the implementer catch all errors themselves and convert to Deny (or whatever).  Just thinking from a security stand point we may want to enforce "If you couldn't answer the question.. then the answer is no").

Just a thought.

mmacfadden (approved):

dnorman:
> How does this interplay with the `client`? A lot of the current methods pass in the client, but that is really to get the node. So just thinking through what the code would look like. There is perhaps some room to rethink the react context provider. I would have to have to do something like this in each method.
> 
> ```ts
> const client = useClient();
> const my_creds = ...
> const context = get_context(my_creds);
> 
> // use the context.
> ```
> 
> Seems like most of the time, we don't really want the client. We are just using the client to get the node. So perhaps initializing a context somewhere else in the app (e.g. with a provider) and doing something like the below might reduce builder plate.
> 
> ```ts
> const context = useContext();
> const artists = await Artist.fetch(context, "status = 'Active'");
> ```

Bingo. This is exactly the idea. useClient goes away, and is replaced by useContext(). 

The client's job is to keep the node connected to the server node. That can be in the background.
It was done this way before purely as an expedience.

The Node has a Generic parameter, which isn't compatible with wasm-bindgen - so that can't be exposed to Typescript
But the Node isn't really useful without Context Data that matches what the PolicyAgent expects.

The next PR will have a `Context` struct, which will have no exposed generics externally, but internally will contain a trait object containing a reference to the Node, and the typed ContextData which matches the Node's PolicyAgent associated type.

So, the rust code will look like

```
#[wasm_bindgen]
pub struct MyCreds{...}

pub struct MyPolicyAgent{..}
impl PolicyAgent for MyPolicyAgent{
   type ContextData = MyCreds;
   ...
}

fn get_context(data: MyCreds) -> Context { // Context is wasm-bindgen

    // Node is instantiated inside the wasm_bindgen(begin) fn
    let node : Node<MyPolicyAgent> = get_global_node();
   
    node.context(data)
}
```

Then in the TS, the UserAuth service would have something like 

```
let context : Context | null = null;
class UserAuth{
    context: Context | null;
    ...
    userLoggedIn(whatever_token){
       ...
       let myCreds = MyCreds::new(...);
       context = get_context(myCreds);
    }
} 

function useContext() {
   return context;
}

```

and the component would have something like :

```
function AlbumListComponent (artist: ArtistView){
    let context = useContext();
    let albums = Album.subscribe(context, {artist}); // pending FromWasmABI magic for View -> edge predicate casting
    
    return (
        <div>
        { results.map((album) => <AlbumComponent album>) }
        </div>
    )
}
```

There is an option to push this context into a global in the rust code so we don't strictly even need the context argument to Album subscribe. Lets discuss that. At a minimum, we will need it for transactions

```
function incrementSales(album: AlbumView, newSales: number){
    let context = useContext();
    let trx = context.begin();
    ...

   // OR
     context.transact((trx) => {
        album.edit(trx).sales.increment(newSales)
     });
    
}
```

core/src/policy.rs
dnorman: Yeah, TBH I think the PolicyAgent code is going to change a lot. The main thing this PR achieves is just to break down the type system changes so the Jenga tower doesn't fall over.

I'm going to merge this PR, not because it's ready, but because it's not harmful. This does put the main branch into a non-releasable state, which I don't love, but it's just because its a breaking API change, not because it actually breaks the happy path.

Issue #55, opened 2025-02-15
Subscriptions in Rust not getting called


mmacfadden:
When creating a subscription in Rust. It seems like the `ItemChange::Initial` is getting triggered but not the other events.

Here is some code that reproduces the issue:

```rust
use std::sync::Arc;

use ankurah::Model;
use serde::{Deserialize, Serialize};
use bb8_postgres::PostgresConnectionManager;
use tokio;

use ankurah::{changes::ChangeSet, Node};
use ankurah_storage_postgres::Postgres;
use tracing::Level;

#[derive(Model, Clone, Serialize, Deserialize)]
pub struct User {
    pub username: String,
    pub first_name: String,
    pub last_name: String,
    pub email: String,
    pub status: String,
}

#[tokio::main]
async fn main() -> anyhow::Result<()> {
    tracing_subscriber::fmt().with_max_level(Level::INFO).init();

    // Initialize Ankurah
    let manager = PostgresConnectionManager::new_from_stringlike(
        "host=localhost port=5432 user=postgres password=postgres dbname=ankurah-test",
        tokio_postgres::NoTls,
    )?;
    let pool = bb8::Pool::builder().build(manager).await?;

    let storage_engine = Postgres::new(pool)?;
    let node = Node::new_durable(Arc::new(storage_engine));

    println!("Ankurah Node Created");

    // Create an initial user that should match the subscription.
    // This should trigger the Initial change event.
    println!("Creating tfox");

    let tx1 = node.begin();
    let tfox_id = tx1.create(&User {
        username: "tfox".to_owned(),
        first_name: "tom".to_owned(),
        last_name: "fox".to_owned(),
        email: "tfox@example.org".to_owned(),
        status: "active".to_owned()
    }).await.id().clone();
    tx1.commit().await?;


    // Subscribe.  The Initial event will be printed.
    node.subscribe(
        "status = 'active'", 
        move |changeset: ChangeSet<UserView>| {
            for change in changeset.changes {
                match change {
                    ankurah::changes::ItemChange::Initial { item } =>  {
                        println!("Initial: {}", item.username());
                    },
                    ankurah::changes::ItemChange::Add { item, ..} =>  {
                        println!("Add: {}", item.username());
                    },
                    ankurah::changes::ItemChange::Update { item, .. } => {
                        println!("Update: {}", item.username());
                    },
                    ankurah::changes::ItemChange::Remove { item, .. } => {
                        println!("Remove: {}", item.username());
                    }
                }
[...]

PR #57, opened 2025-02-17
repro test for #55

dnorman:
Still investigating the cause. Filing PR as draft to contain the RT


core/src/node.rs
```diff
@@ -624,9 +624,9 @@ where PA: PolicyAgent + Send + Sync + 'static
     pub fn get_durable_peers(&self) -> Vec<proto::NodeId> { self.durable_peers.iter().map(|id| id.clone()).collect() }
 }
 
-impl<PA> Drop for Node<PA> {
+impl<PA> Drop for NodeInner<PA> {
     fn drop(&mut self) {
-        info!("Node {} dropped", self.id);
+        info!("NodeInner {} dropped", self.id);
```
dnorman, line 629: logging cleanup

PR #60, opened 2025-02-25
Intermediate type

Aceeri:


core/src/property/mod.rs
```diff
@@ -2,8 +2,80 @@ pub mod backend;
...
 
+use serde::{Deserialize, Serialize};
+
 pub type PropertyName = String;
+
+#[derive(Serialize, Deserialize, Debug, Clone)]
+pub enum PropertyValue {
+    // Numbers
+    I16(Option<i16>),
+    I32(Option<i32>),
```
dnorman, line 19: It strikes me as odd that the Options are internal here. Wouldn't it be more sensible to use Option<PropertyValue> rather than this?

core/src/property/backend/lww.rs
```diff
@@ -119,7 +111,9 @@ impl PropertyBackend for LWWBackend {
     }
 
     fn get_property_value_string(&self, property_name: &str) -> Option<String> {
-        self.values.read().unwrap().get(property_name).map(|v| String::from_utf8_lossy(v).to_string())
+        //let values = self.values.read().unwrap();
+        //values.get(property_name).map(|v| String::from_utf8_lossy(v).to_string())
+        unimplemented!()
```
dnorman, line 116: Yes, this can just cast everything to a string for now. Lets open a separate case for typed Predicate values

core/src/property/backend/lww.rs
```diff
@@ -62,36 +56,34 @@ impl PropertyBackend for LWWBackend {
...
-        let values = self.values.read().unwrap();
-
-        let mut map = BTreeMap::<&PropertyName, &Vec<u8>>::default();
-        for (property_name, value) in &*values {
-            map.insert(property_name, value);
-        }
-
-        let serialized_diff = bincode::serialize(&map)?;
+        let property_values = self.property_values();
+        let serialized_diff = bincode::serialize(&property_values)?;
```
dnorman, line 86: We really want to be storing deltas here of some kind.

For now I suggest we at least find a way to version this so we don't have to do migrations when we later change this to a delta format

Bincode kiiind of isn't the right thing for this, but we can probably get away with it if we wrap it in a Version enum

eg
```
enum Version{
   V1(Whatever)
   // V2 add this later
 }
```

It's not great, but it should work as a band-aid
https://users.rust-lang.org/t/is-bincode-compatible-with-appending-enum-variants/40084

Aceeri: I created a
```rust
struct LWWDiff {
	version: u8,
	data: Vec<u8>,
}
```
instead of the enum just because I don't think serde/bincode guarantees anything with the enum additions. But should work the same way.

storage/postgres/src/lib.rs
```diff
@@ -224,21 +264,6 @@ impl StorageCollection for PostgresBucket {
                             return self.set_state(id, state).await; // retry
                         }
                     }
-                    ErrorKind::UndefinedColumn { table, column } => {
```
dnorman, line 227: @Aceeri At some point I think we should still detect this error, but just not use the field name from it - we can just use it as a prompt to reload the column cache.

Its not urgent, because this would only happen if a column got removed, but it's a recoverable error, so we should probably handle it.

At a minimum we can invalidate the column cache, even if we don't retry immediately. that way the next attempt can succeed

tests/tests/property_backends.rs
```diff
@@ -22,17 +19,41 @@ pub enum Visibility {
     Private,
 }
 
+impl Property for Visibility {
+    fn into_value(&self) -> Result<Option<PropertyValue>, PropertyError> {
```
dnorman, line 23: Consider switching this to use strum Display/EnumString to save the boilerplate https://crates.io/crates/strum

dnorman (approved):
@Aceeri please see comments for possible followup items. Going to approve and merge this as-is.

tests/tests/property_backends.rs
```diff
@@ -110,3 +140,118 @@ async fn pg_property_backends() -> Result<()> {
 
     Ok(())
 }
+
+///
+///
+///
+
+impl ::ankurah::model::Model for Video {
```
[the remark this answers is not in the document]
dnorman: @Aceeri LMK if this makes sense to switch over to the macro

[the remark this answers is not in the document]
Aceeri: Oh it does, it should be removed now. That was just for testing purposes.

PR #91, opened 2025-04-18
Big changes - Mainly about events

dnorman:
Event upgrades
    
    * Event ids are now sha256 hashes of their content
    * Ordering determinations based on event lineage traversal
    * Metadata catalog to store key metadata
    * System root creation and joining for automatic resetting of ephemeral node state
    * Next phase of PolicyAgent, including Attestations for memorializing policy decisions
    * Misc performance enhancements and tweaks

I pulled on a lot of threads in this branch, and kind of unraveled a good part of the sweater.
I generally don't care for bit monolithic commits like this, but it's probably for the better in this case.


dnorman:
Just looking at this thing sideways is going to create conflicts. Merging!

PR #93, opened 2025-04-24
Renamed the "POJO" class and added another create utility method.

mmacfadden:


dnorman (approved):
Looks good! Good call on doing create_one now

PR #120, opened 2025-07-18
Predicate macro

dnorman:
# Add AnkQL Placeholders and Compile-Time Query Macros

This PR introduces two complementary query building approaches for Ankurah:

## Runtime Placeholders
AnkQL now supports `?` placeholders for safe parameterized queries:
```rust
let predicate = parse_selection("name = ? AND age > ?")?;
let populated = predicate.populate(vec!["Alice", 25])?;
```

## Compile-Time Macros
Three macros that compile AnkQL queries directly to AST at build time:

**predicate!** - Core query building with automatic placeholder expansion:
```rust
// Automatic expansion: {name} becomes name = {name} when grammar expects comparison
let pred = predicate!({name} AND {>age});  // name = {name} AND age > {age}

// List expansion for IN clauses
let statuses = vec!["active", "pending"];
let pred = predicate!(status IN {statuses});

// Quoted syntax - necessary for quoted string literals and positional arguments:
let pred = predicate!("status = 'active' AND {name}");  // name = {name}
let pred = predicate!("name = {} AND age > {}", name, age);
```

**fetch!** - convenience wrapper for fetch:
```rust
let users = fetch!(ctx, {name} AND {>age}).await?;
// Equivalent to: ctx.fetch(predicate!({name} AND {>age})).await?
```

**subscribe!** - convenience wrapper for subscribe:
```rust
let handle = subscribe!(ctx, callback, "{user_id} AND status = 'active'").await?;
// Equivalent to: ctx.subscribe(predicate!({user_id} AND status = 'active'), callback).await?
```

**Benefits:** Compile-time query syntax validation (but not yet comparison data type validation), zero runtime parsing, SQL injection protection, ergonomic syntax.


core/src/collation.rs
```diff
@@ -65,6 +65,7 @@ impl Collatable for ast::Literal {
                 bits.to_be_bytes().to_vec()
             }
             ast::Literal::Boolean(b) => vec![*b as u8],
+            ast::Literal::Placeholder => panic!("Placeholders should be replaced before collation"),
```
mmacfadden, line 68: I am not in love with the panic(s) here.  Probably fine for now I suppose.

mmacfadden (approved):

core/src/collation.rs
dnorman: Replaced almost all of the panic/unimplemented invocations, except for one in reactor which I will remove later.

PR #125, opened 2025-07-29
Add version bump workflow

dnorman:
Adds a GitHub Action that automatically bumps crate versions when the `RELEASES` file is modified in a PR.

### What it does:
- **`.release/bump-version.sh`** - Reads the top line of `RELEASES` and runs `cargo release version` to update `Cargo.toml` files
- **`.github/workflows/version-bump.yml`** - Triggers the script when `RELEASES` is changed in PRs

### Behavior:
- Exits early if the version tag already exists (already published)
- Exits early if crate versions are already current (no work needed) 
- Only commits `Cargo.toml` changes when version bumping actually occurs
- Comments on PR when version bump is performed

This moves version bumping from post-merge (in publish workflow) to pre-merge (in PRs), making version changes reviewable and avoiding the need to bypass branch protection rules.

### DIFF
--- a/.github/workflows/version-bump.yml
+++ b/.github/workflows/version-bump.yml
@@ -1,0 +1,81 @@
+name: Auto Version Bump
+
+on:
+  pull_request:
+    paths:
+      - "RELEASES"
+    types: [opened, synchronize]
+
+permissions:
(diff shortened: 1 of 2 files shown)


dnorman:
Decided to undo this approach. Way slower than expected

PR #132, opened 2025-08-14
EntityView and Active Values now implement Signal traits

dnorman:
This PR implements the ankurah_signals `Signal` and `Subscribe` traits for Views and Active values

Please do note that at present these are only receiving local values.
This PR does not keep remote subscriptions active after the active query subscription has been dropped. That will be coming soon in a subsequent PR.

Basic Subscribe example 
```
    let album = context.get::<AlbumView>(album_id).await?;
    
    //  create individual subscriptions to the AlbumView, and it's two fields
    let _h1 = album.subscribe(w);
    let _h2 = album.name().subscribe(w2);
    let _h3 = album.year().subscribe(w3);
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -95,6 +95,7 @@ dependencies = [
  "ankql",
  "ankurah-derive",
  "ankurah-proto",
+ "ankurah-signals",
  "anyhow",
  "append-only-vec",
  "async-trait",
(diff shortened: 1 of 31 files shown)


dnorman:
Reviewed stem to stern. LGTM! lol :)

PR #163, opened 2025-10-21
Fix `LiveQuery::Signal` to fire on entity changes, not just membership changes

dnorman:
## Background
In 0.6.0, `LiveQuery` was introduced with a `Signal` implementation that delegated to `ResultSet`.

Unfortunately, the `ResultSet` Signal implementation only notifies of changes to set membership, and not changes to the entities within that set.

Because Entity Views are themselves Signals which can be discretely observed for changes, we should have a method of observing only set membership changes and omitting notifications for non-membership-affecting Entity changes. In such cases, the commensurate UI list component (table, etc.) would use `view.resultset.get()` (`.items` in JS/TS), and its Observer would bind to the `ResultSet` signal and receive only membership changes.

The Item UI component would then bind to each Entity View directly to be notified of content changes at the entity level. For example, when the Item component calls `item.name()` (or the `.name` accessor in TS/JS) during its first render cycle, it would automatically cause that Item component's Observer to track the item Entity View for discrete changes. This is how things previously worked in 0.6.x.

**However**, in cases where the item count is particularly high, it may not be efficient to have hundreds or thousands of individual components tracking individual entity view signals with individual Observers. Or it may be impractical to have rows as components rendered within the parent component's render context (e.g., antd's Table component). In these cases, there is no option to observe entity views individually, and we must re-render the table component as a whole.

### Solution

Both the `Signal` and `Subscribe<ChangeSet<R>>` trait implementations for `LiveQuery` now delegate to the underlying `ReactorSubscription` broadcast, which fires on all entity changes:
- `Signal` implementation provides notification-only semantics (object-safe: `Arc<dyn Fn(()) + Send + Sync>`)  
- `Subscribe` implementation provides stateless conversion of `ReactorUpdate` → `ChangeSet<R>`

To support this efficiently, `broadcast::Listener<T>` was converted to an enum:
- `Payload(Arc<dyn Fn(T) + Send + Sync>)` - receives the broadcast value
- `NotifyOnly(Arc<dyn Fn() + Send + Sync>)` - notification only, ignores payload

This avoids unnecessary cloning of `ReactorUpdate` payloads when `Signal` listeners don't need the data—they're just tracking "something changed" for reactive re-evaluation.

Types renamed to `BroadcastListener` and `IntoBroadcastListener` to distinguish from `signal::Listener` and `subscribe` listener types.

## Also Included in this PR - ReactObserver Diagnostics

Added `useObserveDiag()` hook for debug-time profiling:

**API:**
- `useObserveDiag(name: string)` - sets observer name and returns the current observer handle
- `observer.signalCount: number` - getter for count of signals currently tracked
- `observer.name: string | undefined` - getter for the observer name

**Build behavior:**
- Debug builds: full functionality for diagnostics
- Release builds: not exported, zero overhead

**Usage:**
```typescript
const observer = useObserve();
try {
  const obs = useObserveDiag("SessionsTable");
  const items = query.get();
  console.log(`${obs.name} tracking ${obs.signalCount} signals`); // query.get() results in 1 signal tracked
  return <div>...</div>;
} finally {
  observer.finish();
}
```

## Bug Fix: Concurrent Events Not Persisted

Fixed a critical bug where concurrent events received via `CommitTransaction` were applied to entity heads but never persisted to storage, causing "Events not found" errors.

**Root cause:** `apply_event()` returned `false` for concurrent events despite modifying the entity head, causing `commit_remote_transaction` to skip event/state persistence.

**Fix:** Changed `apply_event()` to return `true` for concurrent events (line 283 in entity.rs).

**Note:** The underlying `NotDescends` concurrency handling remains incomplete and should be addressed per `concurrent-updates/spec.md`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -37,7 +37,7 @@ dependencies = [
 
 [[package]]
 name = "ankql"
-version = "0.6.2"
+version = "0.6.3"
 dependencies = [
  "anyhow",
  "base64 0.22.1",
@@ -54,7 +54,7 @@ dependencies = [
 
 [[package]]
 name = "ankurah"
-version = "0.6.2"
+version = "0.6.3"
 dependencies = [
  "ankql",
  "ankurah-connector-local-process",
@@ -75,7 +75,7 @@ dependencies = [
 
 [[package]]
 name = "ankurah-connector-local-process"
-version = "0.6.2"
+version = "0.6.3"
 dependencies = [
  "ankurah-core",
  "ankurah-proto",
@@ -87,7 +87,7 @@ dependencies = [
 
 [[package]]
 name = "ankurah-core"
-version = "0.6.2"
+version = "0.6.3"
 dependencies = [
  "ankql",
  "ankurah-derive",
@@ -123,7 +123,7 @@ dependencies = [
 
 [[package]]
 name = "ankurah-derive"
-version = "0.6.2"
+version = "0.6.3"
 dependencies = [
  "ankql",
  "maplit",
@@ -138,7 +138,7 @@ dependencies = [
 
 [[package]]
 name = "ankurah-example-server"
-version = "0.6.2"
+version = "0.6.3"
 dependencies = [
  "ankurah",
  "ankurah-storage-sled",
@@ -154,7 +154,7 @@ dependencies = [
 
 [[package]]
 name = "ankurah-proto"
-version = "0.6.2"
+version = "0.6.3"
 dependencies = [
  "ankql",
  "anyhow",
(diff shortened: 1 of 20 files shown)


signals/src/signal.rs
```diff
@@ -7,14 +7,32 @@ pub use map::*;
 pub use mutable::*;
 pub use read::*;
 
+pub type Listener = std::sync::Arc<dyn Fn(()) + Send + Sync + 'static>;
+
+/// Wraps any broadcast::ListenerGuard<T> in a type-erased container
+pub struct ListenerGuard(Box<dyn crate::broadcast::TListenerGuard + Send + Sync + 'static>);
```
dnorman: Previously, we were using broadcast::ListenerGuard<()> for all signals, which limited which broadcasts we could drive signals with to only Broadcast<()>

This didn't work for ReactorSubscription, because it uses a Broadcast<ReactorUpdate> - and we need to drive the LiveQuery Signal impl off of the ReactorSubscription in order to get content updates

PR #179, opened 2025-12-15
Fix concurrent DDL race conditions in Postgres storage

dnorman:
**Problem:** Intermittent `DDL error: db error` failures and bb8 connection pool timeouts during startup when multiple live queries trigger collection initialization simultaneously.

**Root Cause:** PostgreSQL DDL operations (`CREATE TABLE`, `ALTER TABLE ADD COLUMN`) require exclusive locks and can deadlock when executed concurrently. Multiple tasks were racing into `collection()` initialization.

**Solution:**
- Wrap DDL operations with PostgreSQL advisory locks to serialize schema changes per-collection
- Improve error logging to show full `DbError` details instead of opaque "db error"

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -37,7 +37,7 @@ dependencies = [
 
 [[package]]
 name = "ankql"
-version = "0.7.4"
+version = "0.7.5"
 dependencies = [
  "anyhow",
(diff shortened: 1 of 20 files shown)


mmacfadden (approved):
LGTM

Issue #85, opened 2025-04-11
Property registration


dnorman:
For a variety of reasons, we need metadata about Model properties.

At present we're being fast and loose with property names, which causes issues with name/type collisions.
It makes type casting difficult for migrations, and it makes predicate parsing difficult, because we don't know what's a property and what's not.

Acceptance criteria:

1. Property definitions should be represented by entities in the _ankurah_property collection
2. Entity definitions should by default be upserted on the basis of collection name + property name + type identifier. (at read/write accessor usage time?)
3. Entity lookup is built and maintained using a subscription
4. ankql::ask::Identifier includes property id (entity id)
5. predicate building for Properties which have not been defined by a model will fail

Stretch goal:
Property renames should be achievable by expressly referencing the property entity id on the model

dnorman:
The #289 RFC draft engaging this issue's acceptance criteria is up: https://github.com/ankurah/ankurah/blob/model-property-metadata/specs/model-property-metadata/rfc.md (decision summary on #289). Scorecard for this issue: AC1 satisfied as written (`_ankurah_property` as a dedicated collection), AC3 and AC5 satisfied, AC2 renegotiated on trigger granularity (per model at first use) and read-path timing (cache-only until first mutating use), AC4 via a new resolved `Identifier` node carrying the property entity id, stretch goal via a derivation anchor attribute rather than a literal entity id (ids are system-root-scoped, so a source-embedded id would pin a binary to one system). Renegotiation arguments are in RFC sections 5.2, 5.3, 5.8 and collected in section 10.

PR #236, opened 2026-01-30
Fix empty string missing property via property registration

### LINKED ISSUE #175: Empty string results in missing properties
Ankurah: 0.7.2

I have a user model similar to this:

```#[derive(Model, Clone, Serialize, Deserialize)]
pub struct User {
    pub username: String,

    pub first_name: String,

    pub last_name: String,

    pub email: String,

    pub phone_number: String,
}
```

When creating users the `title` property was getting set to an empty string.  So every user had an empty string value.

```
let user_entity = trx
    .create(&User {
        first_name: user.first_name.clone(),
        last_name: user.last_name.clone(),
        username: user.username.clone(),
        email: user.email.clone(),
        title: "".to_string(),        // This Property
    })
```

When looking in PostgreSQL, the `title` column was not created.  On the consuming side, accessing the `title()` field generates a "missing property" error.  Changing the code to the below to actually set the title resolves the error.

```
let user_entity = trx
     .create(&User {
        first_name: user.first_name.clone(),
        last_name: user.last_name.clone(),
        username: user.username.clone(),
        email: user.email.clone(),
        title: user.title.clone().unwrap_or_default(),
    })
```

It seems like setting the value to an empty string might not trigger property creation.

dnorman:
## Summary

Fixes #175 - Empty strings are treated as missing properties because Yrs CRDTs don't persist empty text operations.

This PR implements property registration (#85) as the solution:
- Models and Properties become entities in the system collection
- Backends key by property entity ID instead of string names
- Read path returns defaults for missing required properties

## Approach

- **Two-phase registration**: Models first, then Properties (handles Ref<T> references)
- **Language-agnostic types**: Backend kind + value type descriptors
- **Registration trigger**: On first `trx.create::<Model>()`

## Status

- [x] Spec and plan documented
- [ ] Expand sys::Item enum
- [ ] Backend refactoring for property ID keying
- [ ] Registration flow implementation
- [ ] Read path updates with empty string fix

## Test plan

- [ ] Unit tests for property registration
- [ ] Integration test: create entity with empty string, reload, verify no error
- [ ] Test optional vs required property behavior

### DIFF
--- a/core/src/value/mod.rs
+++ b/core/src/value/mod.rs
@@ -168,6 +168,41 @@ impl ValueType {
     }
 }
 
+// Conversions between core::ValueType and proto::sys::ValueType
+impl From<proto::sys::ValueType> for ValueType {
+    fn from(v: proto::sys::ValueType) -> Self {
+        match v {
+            proto::sys::ValueType::I16 => ValueType::I16,
+            proto::sys::ValueType::I32 => ValueType::I32,
+            proto::sys::ValueType::I64 => ValueType::I64,
+            proto::sys::ValueType::F64 => ValueType::F64,
+            proto::sys::ValueType::Bool => ValueType::Bool,
+            proto::sys::ValueType::String => ValueType::String,
+            proto::sys::ValueType::EntityId => ValueType::EntityId,
+            proto::sys::ValueType::Object => ValueType::Object,
+            proto::sys::ValueType::Binary => ValueType::Binary,
+            proto::sys::ValueType::Json => ValueType::Json,
+        }
+    }
+}
+
+impl From<ValueType> for proto::sys::ValueType {
+    fn from(v: ValueType) -> Self {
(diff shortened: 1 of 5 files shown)


dnorman:
Superseded by the #289 RFC (specs/model-property-metadata/rfc.md on branch model-property-metadata, ratified 2026-07-05). The RFC absorbs this PR's spec decisions with attribution (entity shape, language-agnostic type descriptors, read-path defaulting, Model-is-not-Collection) and diverges on catalog shape (dedicated collections instead of sys::Item variants), backend identity (open registry name strings instead of a closed BackendKind enum, per #267), and registration trigger (model first-use instead of first create). Its deferred entity-id determinism question is resolved in the RFC by derived ids with frozen genesis encoding. Phase A will be implemented fresh; #175 closure rides Phase A.

PR #252, opened 2026-03-10
Enable offline-first for ephemeral nodes

dnorman:
## Summary
- Ephemeral nodes with a cached system root from a previous session now mark `system_ready` immediately on load, enabling offline operation
- When they reconnect, `join_system` verifies the root and hard_resets if it has changed
- Previously, ephemeral nodes always waited for the server even when they had valid cached data

## Test plan
- [x] All existing system tests updated and passing
- [x] Full test suite passes (no regressions)
- [ ] Verify ephemeral node works offline with cached data
- [ ] Verify ephemeral node correctly re-joins when root changes on reconnect

### DIFF
--- a/README.md
+++ b/README.md
@@ -138,7 +138,7 @@ server.run(&quot;0.0.0.0:9797&quot;).await?;</code></pre>
 <pre><code transclude="docs/example/server/src/main.rs#rust-client-example">let storage = SledStorageEngine::new_test()?;
 let node = Node::new(Arc::new(storage), PermissiveAgent::new());
 let _client = WebsocketClient::new(node.clone(), &quot;ws://localhost:9797&quot;).await?;
-node.system.wait_system_ready().await;
+node.system.wait_system_ready().await; // Wait until a local system root is available
 
 // Create album
 let ctx = node.context(ankurah::policy::DEFAULT_CONTEXT)?;
@@ -150,7 +150,7 @@ trx.commit().await?;</code></pre>
 <pre><code transclude="docs/example/wasm-bindings/src/lib.rs#client-example">let storage = IndexedDBStorageEngine::open(&quot;myapp&quot;).await?;
 let node = Node::new(Arc::new(storage), PermissiveAgent::new());
 let _client = WebsocketClient::new(node.clone(), server_url)?;
-node.system.wait_system_ready().await;
+node.system.wait_system_ready().await; // Wait until a local system root is available
 
 let context = node.context(DEFAULT_CONTEXT)?;
 let _albums = context.query::&lt;ankurah_doc_example_model::AlbumView&gt;(&quot;year &gt; 1985&quot;)?;</code></pre>
--- a/ankurah/src/lib.rs
+++ b/ankurah/src/lib.rs
@@ -93,7 +93,7 @@
 //!
 //!     // Connect nodes using local process connection
 //!     let _conn = LocalProcessConnection::new(&server, &client).await?;
-//!     client.system.wait_system_ready().await; // Wait for the client to join the server "system"
+//!     client.system.wait_system_ready().await; // Wait until the client has a usable local system root
 //!
 //!     // Get contexts for the server and client
 //!     let server = server.context(ankurah::policy::DEFAULT_CONTEXT)?;
--- a/core/src/context.rs
+++ b/core/src/context.rs
@@ -175,6 +175,26 @@ where
     SE: StorageEngine + Send + Sync + 'static,
     PA: PolicyAgent + Send + Sync + 'static,
 {
+    fn should_fallback_to_local(cached: bool, error: &RetrievalError) -> bool {
+        if !cached {
+            return false;
+        }
+
+        match error {
+            RetrievalError::NoDurablePeers => true,
+            RetrievalError::RequestError(req_err) => match req_err {
+                crate::error::RequestError::PeerNotConnected => true,
+                crate::error::RequestError::ConnectionLost => true,
+                crate::error::RequestError::SendError(_) => true,
+                crate::error::RequestError::InternalChannelClosed => true,
+                crate::error::RequestError::ServerError(_) => false,
+                crate::error::RequestError::UnexpectedResponse(_) => false,
+                crate::error::RequestError::AccessDenied(_) => false,
+            },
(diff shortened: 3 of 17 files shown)


dnorman:
Review notes from a verified multi-agent pass over this PR (every finding below was adversarially confirmed against the branch, not just pattern-matched). Also one sequencing note up top.

## This PR now solely owns the `system_ready` gate-flip

The jwt-auth branch (#253) previously carried a copy of the `if has_root && self.0.durable` → `if has_root` change; it was removed there (`ffd247aa`) and deferred here, because the flip is inseparable from an unanswered write-path question:

- **Semantic change:** an ephemeral node restarting with a cached root can now create contexts — and therefore commit — before/without ever connecting. Previously `context()` failed and `wait_system_ready()` blocked until `join_system` completed against a live durable peer at least once per process.
- **Stranded offline writes:** a commit with zero connected durable peers persists locally, `relay_to_required_peers` iterates an empty peer list and returns `Ok`, and there is no outbox or resend-on-reconnect anywhere — the write never reaches the durable node, even after a clean reconnect with a matching root. The spec's Scope and Deferred Work sections don't mention writes at all.
- **Race for existing apps:** the documented startup pattern (`WebsocketClient::new` → `wait_system_ready().await` → write) now resolves the wait instantly on restart, racing the websocket handshake — an app can strand a write in the window before the peer registers. This was structurally impossible before.
- The related ready-before-`hard_reset` race (stale cached root + async reset wiping a just-committed write) is real but **dev-only by design** — root mismatch doesn't occur in production.

**Ask:** scope offline writes explicitly in the spec (even as "unsupported: commits fail loudly with no durable peer"), or add the outbox. Silent local-only success is the worst of the options.

## Confirmed findings (most severe first)

1. **`core/src/node.rs` (Node::new flush loop) — panics on wasm.** `tokio::time::sleep` runs under `wasm_bindgen_futures::spawn_local` with no tokio reactor; the loop panics on its first poll in every browser `Node::new`. In-crate fix pattern: `futures_timer::Delay` (already used in `client_relay.rs:422`).
2. **`core/src/node.rs` (UnsubscribeEntities handler) — ranges applied collection-blind to a shared, provenance-less set.** The handler discards `collection`, and the per-peer `entity_subscriptions: BTreeSet` spans all collections *and* is also populated by query processing (`subscription_state.rs:295`). ULIDs are timestamp-ordered, so a range over two dead album ids commonly contains a live pet id (or a query-subscribed id) → server silently strips a live subscription; client never learns. Suggest: plain `Vec<EntityId>` instead of ranges (they only compress runs of consecutive dead ids), per-collection scoping, and separating get-driven from query-driven ids.
3. **`core/src/context.rs` (get_entity) — resident-entity fast path bypasses policy.** Returns any in-memory entity with zero policy calls, where main's ephemeral get always hit the durable peer whose Get handler enforces `check_read` per context. Context A's authorized get makes X resident; a context the server would deny then reads X locally. Moot for PermissiveAgent; directly undermines #253's JwtAgent. **On rebase over #253, this fast path must pick up #253's `check_read` (added to this same function there).**
4. **`core/src/node.rs` — client/server subscription state diverges permanently (4 confirmed mechanisms).** (i) `ensure_entity_subscription` inserts into `tracked` before the request; failure only warns and the insert-guard suppresses all retries → resident entity silently stale while connected. (ii) `flush_dead_entity_subscriptions` deletes tracked entries *before* sending, and one `send_message?` failure aborts remaining collections and peers → server-side leak with no retry path. (iii/iv) Nothing sequences the spawned subscribe against the flush's send → drop-revive races end subscribed-but-untracked or unsubscribed-but-tracked. The spec's own pending-set model (clear only after successful send) fixes ii–iv.
5. **`core/src/node.rs` (resubscribe_tracked_entities_to_peer) — first error abandons all remaining entities, deterministically.** `?` inside the per-entity loop, warn-only caller, failed entities stay tracked. One policy-denied entity re-aborts at the same BTreeMap position on every reconnect; everything sorting after it stays stale until restart. Fix: batch per `(collection, cdata)` (the protocol already takes `Vec<EntityId>`; handler requires exactly one cdata) with per-item error handling.
6. **`core/src/node_applier.rs` — context fallback blends unrelated credentials.** When relay contexts are empty, updates apply under the union of *all* tracked contexts (unfiltered by peer or entity), and lineage `GetEvents` requests carry one AuthData per context — user B's credentials ride on retrievals for entities only A subscribed to. Also resurrects the unsolicited-update path main rejected. Precise fix is already supported: update items carry `(collection, entity_id)` and `tracked` stores cdata per entity — look it up.
7. **`proto/src/request.rs` — version mismatch fails silently (reframed: wire compat across minors is not guaranteed, per maintainer).** The new variants renumber bincode indexes, which is fine policy-wise — but a version-skewed peer today gets warn!-and-drop frames (requests hang forever) or, worse, payloads that decode as the *wrong variant*. Since `Presence` is exchanged at connect time, a protocol version field checked at handshake — refuse loudly on mismatch — converts undefined cross-version behavior into a clean error. Relevant for cached browser wasm bundles vs freshly deployed servers.
8. **Eternal 100ms poll vs the spec's own lifetime model.** The spec mandates refcount semantics (0→1 subscribe, 1→0 enqueue + 100ms debounce, cancel on revival); the implementation polls residency every 100ms forever, even with zero tracked entities — a perpetual wakeup source on battery-powered wasm clients, with no actual debounce-cancel semantics. A drop hook (or parking on a Notify while empty) implements the spec on existing infrastructure.

## Cleanup worth batching into the same revision

`should_fallback_to_local` is a verbatim second copy of client_relay's retryability match — extract `RequestError::is_transient()`; `subscribe_entities_with_peer` is the third hand-rolled copy of the known-heads + apply-deltas flow (the copies already disagree on whether "known head" comes from storage or resident entities); the server's `subscribe_entities` duplicates the Get handler's policy loop and the delta-generation loop (third copy); `SubscriptionHandler::remove_entities` has no callers.

## Suggested split

The reads story (cached fallback, gate-flip for reads, and their tests) is solid and severable. Findings 1–6 and 8 cluster entirely in the entity-subscription protocol. Landing reads + system_ready first (with the offline-write scoping resolved) and revising the entity-subscription protocol as a follow-up would ship most of the value sooner.

Issue #269, opened 2026-07-04
Tracking: concurrency phase 2 (maturation, bulletproofing, then optimization)


dnorman:
Tracking issue for concurrency phase 2. The source of truth is
[`specs/concurrency/phase-2.md`](https://github.com/ankurah/ankurah/blob/concurrent-updates-event-dag/specs/concurrency/phase-2.md)
(that link follows the branch until #201 merges, then the path lives on main); this issue is the
public anchor and status surface, not a duplicate checklist.

**Scope in one paragraph:** mature the concurrency architecture proven in #201.
Workstreams: (A) open-issue triage for situational awareness, (B) literature
review run with Opus subagents, covering compaction/lifecycle, indexed
causality, Merkle-causal systems, merge semantics, deterministic simulation
testing, staleness horizons and divergence bounding (feeding the rejection
horizon decision under a hard convergence requirement), and adjacent database
internals; (C) test bulletproofing: deterministic multi-node simulation
harness, oracle scale-out, backend conformance kit, threat model plus
adversarial wire suite, reactor/LiveQuery coherence, crash/recovery fault
injection, storage engine conformance; (D) the architecture ladder: #268
ingest pipeline, #266 generations plus applied-set, then history lifecycle,
transactional visibility, snapshot authority, validated ingress,
observability, and the #265/#267 remainders; (E) as a separate PR:
benchmarking (harness and baseline land at phase 2 start) then optimization,
including the streaming application consumer.

**Working conventions:** each workstream opens a draft PR immediately at
kickoff so CI and progress are visible from day one; the checklist at the
bottom of the spec is updated in the same commits that land the work;
phase 1 conventions carry over (red-to-green, validation gate, no em dashes).

**Related:** #265, #266, #267, #268 (architecture RFCs; four more to be filed
at kickoff per workstream D), #242-#247, #236, #256, #264, #200 (triage
targets). Phase 1 history: branches `archive/201-concurrent-updates-specs`
and `archive/201-pre-squash`.

dnorman:
Phase 2 kickoff complete (2026-07-04). Status:

Workstreams opened (draft PRs target concurrent-updates-event-dag while #201 is open; they retarget to main after it merges):
- A (issue triage): #277. Triage table appended to the scope doc; every open issue classified and verified complete against the live issue list.
- B (literature review): #275. Seven topic readers running; adversarial synthesis follows. The staleness-horizons topic gates D3 implementation.
[...]

PR #286, opened 2026-07-05
Phase 2 C6: crash and recovery fault injection

dnorman:
Workstream C6 of concurrency phase 2. Tracking: #269. Scope: specs/concurrency/phase-2.md, workstream C item 6.

Draft opened at kickoff per the phase 2 convention. This PR targets the concurrent-updates-event-dag branch while #201 is open; it will be retargeted to main after #201 merges.

Scope:
- A deterministic crash-injection storage wrapper (operation-counter triggered process abort) driven through child-process kill and reopen cycles on real sled directories
- Crash points: between commit_event and set_state, mid-batch, mid-merge, and during entity creation
- Reopen invariants: no persisted state referencing missing events; staged-but-uncommitted events recovered or safely absent; reopened nodes reconverge with peers
- Postgres arm included where the existing container infrastructure reaches; IndexedDB out of scope (no process-kill semantics under wasm test runners)
- Any real defect found lands as an ignored red test plus a filed issue, not an in-PR fix


dnorman:
Review follow-up recorded as #319: the crash harness (already generic over StorageEngine + CrashFlushable) extracts into a shared internal ankurah-testkit crate as C7's opening move, and the per-engine arms (postgres behind its Docker feature, sled) migrate into their engine crates. This PR merges as-is first; the extraction is additive.

Issue #289, opened 2026-07-05
RFC: model and property metadata (defining entity per model and per property)


dnorman:
## Intent (maintainer, 2026-07-05)

Each model gets its own defining entity id, and each property gets its own defining entity id. Model and property metadata become first-class entities in the system rather than compile-time-only facts.

## Why (cases this addresses; connections gathered from the 0.9.0 work, to be validated in the RFC)

- **Property registration cluster (#236, #175):** the empty-string/missing-property bug family is already slated to be fixed "via property registration". Defining entities for properties ARE that registration, made durable and replicated.
- **Schema evolution:** today a property IS its name. With a defining entity, the name becomes a label on a stable identity: renames stop being data migrations, and the (currently unwritten) schema-evolution story gets a real foundation.
- **SQL engines add columns keyed by property name** on demand (postgres/sqlite). Stable property ids give collision-safe, rename-safe column identity.
- **Cross-node strategy agreement:** merge strategy (LWW vs Yrs) is declared per field in code via #[active_type]. Nodes running different model versions can disagree silently. An authoritative per-property strategy on the defining entity makes the merge contract data, not code.
- **Per-field addressing:** field-level signals exist but per-field subscription from a View is unsupported (tests/tests/basic.rs TODO); policy cannot express per-property rules; index config (sled property_config/index_config trees) is engine-local. All three want a stable property identity to hang off.
- **Introspection/tooling:** collections and properties become queryable like any other data.

## RFC questions to work through

- Bootstrapping: metadata entities live in which collection, created when (model first use? explicit migration?), and how do nodes with mismatched model code reconcile against them.
- Relationship to the sysroot/system catalog (core/src/system.rs) and to CollectionId::fixed_name.
- Wire/storage impact: do events reference property ids instead of names, and what does that do to event size and the 0.9 LWW state format (which is versioned: one-byte header, 0xA1 = v1, so a v2 is cheap to introduce).
- Policy hooks: does PolicyAgent gain per-property gates keyed by property entity id.
- Precedent check against phase-2 RFCs #265-#268 and #271-#274 for overlap (esp. #267 backend contract).

Suggested first step: a design doc in specs/ following the phase-2 RFC conventions, then an implementation ladder.

dnorman:
[...]

PR #306, opened 2026-07-05
Protocol version in the Presence handshake (#294)

### LINKED ISSUE #294: Protocol version in the Presence handshake (rolling upgrades)
There is no protocol version in the websocket Presence handshake, so nodes cannot detect version mismatch at connection time. Concretely today: 0.8.x and 0.9.0 nodes can exchange events but not state (0.8.x binaries fail on 0.9 LWW state buffers), and the only guidance is "upgrade all nodes of a system together" (flagged in #284).

- [ ] Add a protocol/format version to Presence (proto/src/message.rs); decide semantics: refuse vs degrade on mismatch
- [ ] Policy hook consideration: should version acceptance be a PolicyAgent decision
- [ ] This becomes load-bearing before the NEXT on-disk or wire format change; sequence it ahead of any format-touching work (e.g. the model/property metadata RFC if it changes event encoding)

dnorman:
Closes #294. Phase 0 of the model/property metadata ladder (#289; RFC at specs/model-property-metadata/rfc.md, rev 3): gates the Phase A wire changes.

## What

- `PROTOCOL_VERSION: u32 = 1` in proto, carried as a new trailing `protocol_version` field on `Presence`. The pre-#294 encoding is a strict byte prefix of the new one (pinned by a proto unit test), so a version-0 (0.9.x or older) handshake is detectable rather than guessed at.
- Semantics: REFUSE on inequality, both directions. Enforced in `Node::register_peer` (now fallible) so every transport inherits it, including local-process which never serializes. Degrade was rejected: RFC rev 3 forbids interim dual encodings, and durable-nodes-upgrade-first is the deployment answer (RFC section 4).
- The comparison lives in `proto::protocol_compatible` so a future version can widen acceptance to a range without touching the handshake again.
- New `Message::PresenceRejected { expected, received }`, sent best-effort before close. The websocket server and clients pre-check the version while the raw sink is at hand so the notice actually flushes; `register_peer` re-enforces for every transport.
- Handshake frames that fail to decode now close the connection with an actionable log (version-0 classification via `proto::is_version0_presence`) instead of dropping the frame and leaving the socket open.

## Deployment note

The 0.9.x boundary with this release is itself unversioned: an old node cannot name the mismatch (it predates the vocabulary) and only observes a close; the new side logs precisely. From this release onward, mismatches are refused cleanly on both ends.

## Deferred

- PolicyAgent hook for version acceptance (the #294 checkbox): stays structural until a concrete use case appears.

## Tests

- proto unit: Presence round-trip, prefix-property pin, version-0 classification, exact-match compatibility.
- tests/tests/protocol_version.rs: register_peer refuses version 0 and version+1 and accepts current; server refuses a version-0 handshake by closing; server refuses v999 with PresenceRejected observed on the wire; client refuses a v999 server, sends PresenceRejected, and never reports connected.
- Validation gate green: core lib, full ankurah-tests, jwt-auth, isolated wasm check, fmt + taplo.


dnorman:
Superseded by #307 (maintainer direction: the Phase A PR is the primary artifact and targets main directly, carrying these two commits at its base). The #294 decision record stands; #294 itself closes when #307 merges.

Issue #305, opened 2026-07-05
Deconflict Collection vs Model: Collection becomes a storage-only concern


dnorman:
Maintainer direction (2026-07-05, #289 review): redefine Collection as a storage-only concern (Collection = table, sort of: where materialized rows live and how requests route), while Model is the data contract (a set of property memberships an entity is anticipated to meet, per the #289 metadata RFC). Today the two terms blur because the collection id is derived from the model name and an entity belongs to exactly one collection.

## Work

- [ ] Terminology ruling written into GLOSSARY.md and the book (coordinate with #283/#291 docs work): Collection = storage placement and routing key; Model = contract; the model entity's `collection` property records the binding as data.
- [ ] API naming audit against the ruling: CollectionId stays the storage/routing key type; Model::collection() reads as "the collection this contract's entities currently materialize into"; flag any surface where "collection" is used to mean schema/contract.
- [ ] sys::Item::Collection disposition. RECOMMENDATION: delete the variant. It has never been constructed (the TODO at core/src/system.rs:87 is dead intent), #[serde(other)] makes removal wire-safe, and the model entity's collection binding covers the mapping. Renaming it to Model was considered and is recommended AGAINST: model definitions are first-class entities in _ankurah_model under the RFC, and a sys::Item::Model variant would reintroduce the sys-catalog conflation the RFC deliberately removed (PR #236's superseded approach). If a storage-only catalog record ("this table exists") is ever needed, add it later under the new terminology.

## Sequencing

The ruling and glossary work can land any time; the variant deletion fits Phase A or immediately after. Related: #289 (RFC section 4a records the trajectory), collection-excision issue (Collection-as-table becomes literal there).

dnorman:
Maintainer nomenclature ruling to encode (2026-07-05): Model = the display/matching name of a model definition entity; Collection = the storage table name. The current code does not make this distinction; it must be created NOW, ahead of entity unification. Forward note for the unification work: once entities are unified, each entity will need to know which models it meets so it can be materialized into the respective collections (entity -> contracts -> materialization targets).

dnorman:
Design discussion notes from the #307 review (2026-07-09), extending this issue's direction. Subject to ratification.

[...]

Issue #304, opened 2026-07-05
Collection excision: move canonical state and head to a shared entities table (postgres/sqlite)


dnorman:
Excluded from the #289 metadata project by decision (2026-07-05); recorded with its target shape in specs/model-property-metadata/rfc.md section 4a. Filed so the deferred storage refactor has a tracking anchor.

## Target shape (maintainer, 2026-07-05)

- Canonical state and head move OUT of the per-model tables into a shared entities table.
- Materialized values REMAIN per model exactly as now: a named postgres/sqlite table per model with its dynamic property columns.
- Whether head is duplicated onto the materialization tables is undecided.
- Sled already has precisely this split and is the reference: global entities_tree for canonical state, per-collection materialized trees for values (storage/sled/src/database.rs:17-20, storage/sled/src/collection.rs:99-133). The excision is, in essence, making postgres/sqlite look like sled.
- Consistent with the identity layer: event identity already excludes the collection (proto/src/data.rs:17), and PR #236's plan recorded the unified-collection intent.

## Current coupling to untangle

Postgres/sqlite interleave canonical and materialization writes in one table today: the state table carries id, state_buffer, head, attestations ALONGSIDE dynamic property columns (storage/postgres/src/lib.rs:260-269), and set_state performs both writes in one pass (lib.rs:342-391; sqlite engine.rs:260-294).

## Constraint already placed on other work

The metadata RFC's Phase C engine work (column creation, catalog-driven rename DDL) runs inside set_state and MUST keep canonical-write vs materialization-write seams clean so this split stays mechanical (RFC 4a). This issue consumes that cleanliness.

## Sequencing

Independent of the metadata phases; coordinate with #270 (durable storage migrations) for the table migration, and with #294 if any wire-visible behavior changes (none expected: this is storage layout only).

Issue #303, opened 2026-07-05
Metadata schema evolution follow-ons: membership tombstones and declared-coercion retypes


dnorman:
Two related follow-ons deferred from the #289 metadata RFC (specs/model-property-metadata/rfc.md sections 4, 5.6, 5.9, 6; ratified 2026-07-05). Both are schema-evolution work that must NOT ride Phase A.

## 1. Membership retirement (tombstones)

When a field leaves a contract, the (model, property) membership entity needs a tombstone flag rather than deletion: the shared property entity may still belong to other contracts, and history referencing the membership remains valid. The membership entity reserves the concept (RFC section 4); this issue defines the semantics: what readers do with a tombstoned membership (projection, predicates, materialized columns), whether tombstones expire, and the interaction with #271's erasure question (GDPR-style property erasure = catalog tombstone plus per-entity scrubbing).

## 2. Declared-coercion retypes (retyped_from)

Under the RFC, a retype mints a NEW property identity; reads are fail-visible via the sibling gate (TypeSkew), and for SHARED properties a retype is a breaking change for every explicit binder by design (RFC 5.9). The escape hatch is a declared coercion: new code declares that property id N supersedes property id O with a cast, so reads resolve through the coercion and binders migrate deliberately. Design questions: where the coercion is declared (membership metadata? a dedicated catalog record?), which casts are legal per (backend, value_type) pair, write-side behavior (write both? write new only?), and when the old id's data is considered migrated.

Blocked on Phase A landing. Seeds for the schema-evolution chapter (#291) and adjacent to phase-3 parking lot item 1 in specs/concurrency/phase-2.md.

dnorman:
The cast-at-read half of this issue now has a concrete design direction (maintainer, 2026-07-05), recorded in specs/model-property-metadata/tasks.md on the metadata branch (PR #307): a sanctioned retype keeps value_type as an identity input (collision safety, #85), mints a new property id, and records a lineage edge (succeeds: old_property_id) REQUIRING old->new castability; reads that miss the new id follow the edge and cast at query time -- the RFC 5.4 sibling gate with recorded permission (today's TypeSkew is exactly this situation without the edge). Optional lazy rewrite-on-save migrates opportunistically. Open design points listed there: normative cast rules (value/cast.rs is the seed), fallible-cast policy, predicate pushdown, interaction with the tombstone half.

dnorman:
[...]

Issue #302, opened 2026-07-05
Catalog-to-binding codegen for explicit id binding (metadata RFC Phase D)


dnorman:
Deferred from the #289 metadata RFC (specs/model-property-metadata/rfc.md section 5.9, ratified 2026-07-05).

## Context

Explicit binding attributes (#[model(id = "...")], #[property(id = "...")]) land in Phase A: parsed, verified fail-closed against the definition's (backend, value_type), membership minted. But entity ids are per-system (all minted ids are root-scoped derivations), so a literal id in source binds that source to one system's catalog. Until tooling exists, hand-authoring ids is awkward for multi-system deployments (dev/staging/prod from one binary).

## What

- [ ] Codegen: emit a binding file (model and property ids for a chosen set of definitions) from a live system's catalog, suitable for build-time include or env-keyed selection per deployment.
- [ ] Print/inspect tooling: resolve a (collection, name, backend, value_type) to its per-system id on demand.
- [ ] Documented practice: anchors (#[property(anchor = "...")]) remain the recommended portable form for single-authored definitions; explicit ids are for shared properties and DDL-authored definitions, provisioned per deployment.

## Open

- Binding file format and location; interaction with the DDL document (a DDL apply could emit the binding file in the same pass).

Depends on Phase A (attributes, catalog). Pairs with the DDL issue; both are the RFC's Phase D.

PR #307, opened 2026-07-06
Metadata catalog: durable-allocated model/property identity and the id-keyed data contract (protocol v3)

### LINKED ISSUE #85: Property registration

### LINKED ISSUE #175: Empty string results in missing properties

### LINKED ISSUE #289: RFC: model and property metadata (defining entity per model and per property)

### LINKED ISSUE #294: Protocol version in the Presence handshake (rolling upgrades)

### LINKED ISSUE #310: StorageEngine::list_collections overrides for postgres/sqlite/indexeddb
list_collections was added additive with a default empty impl (PR #307); sled implements it. The durable catalog warm uses it to avoid materializing empty _ankurah_* trees. Implement the real overrides for postgres, sqlite, and indexeddb-wasm.

### LINKED ISSUE #334: Make durable catalog warming restartable after hard_reset
## Context

This is a correctness follow-up to PR #307. It is intentionally separate and should not expand PR #307 before final review or landing.

PR #307 correctly flushes catalog-derived state during `hard_reset`. This issue completes the durable-node lifecycle after that flush.

## Current behavior

[CatalogManager::start](https://github.com/ankurah/ankurah/blob/4fdc2ad193cdb41a0909637c2210d79cd74bb329/core/src/schema/catalog.rs#L498-L538) starts one warm-and-subscribe task when a durable node is constructed. [CatalogManager::reset](https://github.com/ankurah/ankurah/blob/4fdc2ad193cdb41a0909637c2210d79cd74bb329/core/src/schema/catalog.rs#L807-L828) clears the map, readiness flag, and durable subscription, but it does not start another warm cycle. [SystemManager::hard_reset](https://github.com/ankurah/ankurah/blob/4fdc2ad193cdb41a0909637c2210d79cd74bb329/core/src/system.rs#L228-L274) invokes that reset hook.

A concrete sequence is:

1. Construct a durable node and wait for its initial catalog warm to complete.
2. Register a schema.
3. Call `hard_reset`.
4. Create a new system root on the same node.
5. Attempt registration, selection resolution, or model-ID ingress resolution.

The initial catalog task has already completed. Reset removed its subscription and set catalog readiness to false. No task observes the new system-ready transition, so code waiting in `wait_catalog_ready` can remain blocked.

The existing tests cover reset flushing and catalog readiness separately. The regression requires combining them on a durable node that completed its first warm before reset.

## Proposed direction

Treat durable catalog warming as a restartable, generation-aware lifecycle:

1. Transition the system to not ready.
2. Invalidate the current catalog generation, clear the map, and drop its subscription.
3. Arm exactly one warm cycle for the next system-ready transition.
4. After the new root is ready, scan durable catalog state, install a fresh subscription, and mark the current generation ready.
5. Prevent a pre-reset warm task from installing a subscription or marking a later generation ready.

## Acceptance criteria

- [ ] A durable node that was initially catalog-ready can run `hard_reset`, create a new root, and become catalog-ready again without reconstructing the node.
- [ ] Registration and resolved queries succeed after the reset and recreate sequence within a bounded timeout.
- [ ] Catalog IDs from the old system remain absent after reset.
- [ ] New registrations update the re-warmed map through the fresh durable subscription.
- [ ] Repeated reset and recreate cycles leave exactly one active durable catalog subscription.
- [ ] A delayed pre-reset warm cannot mark the post-reset generation ready.
- [ ] A regression test waits for initial catalog readiness before resetting, so task scheduling cannot mask the behavior.

### LINKED ISSUE #364: ankql integer literals bind as i32 → WrongType on Postgres bigint columns (all i64-field comparisons fail)
Found while making idp's webhook service livequery-driven (synestheticsystems/idp#58, commit 253b6a7): any comparison predicate against an i64 model field fails on the Postgres engine, e.g.

```
ctx.fetch::<AuditEventView>("occurred_at >= 0")
// StorageError(Error { kind: ToSql(0), cause: Some(WrongType { postgres: Int8, rust: "i32" }) })
```

The same predicate works on sled. Since epoch-second timestamps fit in i32 until 2038, effectively **every timestamp comparison** hits this.

Chain (ankurah 0.9.0 from crates.io):
- ankql parses small integer literals as `Literal::I32` (`ankql/src/ast.rs`; `core/src/type_resolver.rs:56` maps `Literal::I32 → ValueType::I32`).
- `storage-postgres/src/sql_builder.rs` binds the param at its literal width; `storage-postgres/src/value.rs` maps `Value::I32 → PGValue::Integer` (int4).
- Materialized i64 model fields are `bigint` (int8) columns; tokio-postgres is strict about integer widths → `WrongType { postgres: Int8, rust: "i32" }`.

The widening machinery already exists — `core/src/value/cast.rs:71` has `(Value::I32, ValueType::I64) → I64` — it just isn't applied to fetch/query predicate literals against the model property's declared type. Suggested fix: resolve literal types against the collection schema (the way `resolve_selection_types` already does for JSON-path comparisons) before parameter binding, or bind all integer literals as i64 when the target column is int8.

Downstream impact: idp's audit-event LiveQuery subscription wanted `occurred_at >= <boot>` to bound the retained resultset over an append-only collection; it ships kind-bounded-only until this is fixed. Repro pinned in idp: `webhooks::service::tests::postgres_predicate_probe` (ignored diagnostic; needs DATABASE_URL).

### LINKED ISSUE #365: LiveQuery: activation failure leaves wait_initialized() hanging forever; error surfaced only at debug level
Found together with #364 (they compose badly): when a durable-node LiveQuery's activation fails, the caller awaiting `wait_initialized()` hangs forever with no visible error.

In `core/src/livequery.rs` (0.9.0), the durable-node construction path spawns activation:

```rust
crate::task::spawn(async move {
    if let Err(e) = inner2.activate(1).await {
        debug!("LiveQuery initialization failed for predicate {}: {}", query_id, e);
        inner2.error.set(Some(e));
    }
});
```

On failure: the error goes to `inner.error` (no public accessor found) and a `debug!` line; `mark_initialized` runs only via the pre-notify hook inside a successful `add_query_and_notify`, so `initialized` is never notified and `wait_initialized().await` (livequery.rs:112) blocks forever.

How this bit in practice: idp's webhook service awaited `wait_initialized()` during server startup; the predicate hit #364 on Postgres, activation failed silently, and the whole server hung before ever listening — no error at info level, nothing to see without `ankurah_core=debug`. (Sled accepted the same predicate, so tests were green.)

Suggestions, any subset:
1. `wait_initialized()` should return `Result` (or a variant like `try_wait_initialized`) that resolves with the stored activation error instead of hanging.
2. Notify `initialized` (or a dedicated failure notify) on activation failure so waiters wake.
3. Log activation failure at `warn!` — a query that will never produce data is not a debug-level event.

There's also a pre-existing `FIXME` on version-awareness in `wait_initialized` (livequery.rs:119) that touches the same code.

Downstream defense meanwhile (idp `253b6a7`): every `wait_initialized()` at startup is wrapped in a 10s `tokio::time::timeout` that fails startup loudly and points at `ankurah_core=debug`.

### LINKED ISSUE #374: ORDER BY canonical-type fallback re-resolves the physical name and is duplicated across sled/idb
Follow-up from the PR #307 final review. The ORDER BY canonical-type fallback re-resolves a physical column name through the display-name index, which is dead in the normal case and wrong-typed in a rename-reuse corner, and it is duplicated across two engines. Branch-introduced.

## Problem

`storage/sled/src/collection.rs` and `storage/indexeddb-wasm/src/collection.rs` both build an ORDER BY key-part type map (`order_types`) keyed by property id before translating to column space, then apply a fallback that re-resolves the already-translated physical column name through the catalog display-name index:

```rust
// after translation, keyed by the physical column name
resolver.resolve(collection, name) ... canonical_value_type(...)
```

Every resolvable ORDER BY item is already present in the pre-translation `order_types` map, so this fallback can only fire for:

- a genuinely unresolvable name, where it returns None anyway, or
- the corner where a property resolves but has no recorded canonical type and its sticky physical column happens to equal another property's current display name (post-rename name reuse), where it returns the other property's type for this column, so the scan collates or errors TypeMismatch on the wrong domain.

The comment claims the fallback keeps historical String collation for unresolvable columns; it does not describe the wrong-type case.

## Fix

Make the pre-translation `order_types` map authoritative (`|name| order_types.get(name).copied()`) and delete the display-name re-resolution. Share the helper between sled and IndexedDB rather than duplicating it. The IndexedDB unit test `resolved_order_type_survives_sticky_physical_name` already exercises the authoritative-map shape.

dnorman:
# Summary

This PR moves Ankurah's model and property identity out of local Rust declarations and into a durable metadata catalog. Models, properties, and model-property memberships receive stable IDs from the durable node. User data, query plans, reactor indexes, and storage engines then use those IDs while keeping collection and field names as mutable display metadata.

It also makes each property's `backend` and `value_type` canonical at first allocation. Later binaries may register a mutually castable local type, but registration never mutates the stored canonical type. Writes, predicate literals, live-query matching, and `ORDER BY` all cast and collate in that canonical domain.

Folded on top of that base are two identity refactors completed on this branch (2026-07-14 through 2026-07-16): the column-layering refactor, which removes the storage concept of a column from the query and identity layers and fixes #374 (after a rename, a query could address the wrong physical column and return wrong rows), and the PropertyKey retirement, which retires the interim write-side key type `core::property::PropertyKey` so that `ankql::ast::PropertyId` is the one durable property identity across backend maps, wire payloads, Yrs root names, and query ASTs. The PropertyKey amendment recorded on #289 is repealed by the retirement; the design documents record the retractions (`specs/model-property-metadata/rfc.md` and `plan.md`). A comprehensive review of the folded work then produced a remediation batch, folded into the final squashed commit.

## Review sequencing

The branch is intentionally squashed into one commit over `main`. The final commit's tree was verified byte-for-byte against the preserved pre-squash tip before the branch was force-pushed; review the PR as one end-to-end change.

Three pieces were carved out as standalone PRs that can merge to main ahead of this one: #379 (the ankql NOT keyword boundary fix), #380 (a constant-false predicate installs no watchers), and #381 (exact protocol-version negotiation). Each was verified independently against main. Once one merges, this branch rebases over it and the corresponding commit drops out as already applied.

## What changed

- Added durable `RegisterSchema` allocation for model, property, and membership definitions.
- Added exact compiled-schema bindings that retain the allocator's model and property IDs, survive display-name renames, and reassert after reconnect. Offline catalog proof may enable local use, but it is not treated as allocator confirmation.
- Changed user state and operations to stable property identity. An earlier revision kept lazy compatibility for legacy name-keyed state; the PropertyKey retirement removed it before merge, so there is no legacy decode and no name fallback anywhere (the version 1 epoch, targeting the unreleased 0.10.0, is not published; development databases reset).
- Changed event, state, delta, and subscription envelopes to carry model IDs. Catalog descriptors travel with the wire envelope so a cold receiver can route data before its cache is warm. The remaining concurrent send-ordering and descriptor-freshness redesign is tracked in #331.
- Added canonical casts and collation to predicate resolution, reactor matching, planners, and engine indexes.
- Added per-engine ID-to-column maps for SQLite, PostgreSQL, IndexedDB, and Sled, including Sled index compatibility checks and rebuilds for pre-epoch collation specs.
- Removed the storage column concept from the query and identity layers (the #374 fix). Resolved references carry `ankql::ast::PropertyId`; every engine refuses an unresolved selection at its fetch boundary (`Selection::check`) and translates identity to a physical column through its own durable map, with no name fallback and a once-only map reload before treating a referenced property as absent. The two wire handlers re-resolve after merging policy predicates, mirroring the local query path.
- Hardened column assignment: a system property whose sanitized name collides with a reserved column or with another identity's column is refused loudly in all four engines (silently absorbing the collision is how two identities could alias onto one column), and suffix exhaustion errors instead of returning a name that is already taken.
- The planner never emits a primary-key index keypart: a leading primary-key `ORDER BY` is absorbed by table-scan direction, and index plans spill it (and every key after it) to the in-memory comparator, which orders entity ids canonically. Previously such a keypart was unresolvable at the engine and permanently broke writes to the collection through a failed index build.
- Added a storage protocol-version record to every engine (`_ankurah_meta` in SQLite, PostgreSQL, Sled, and IndexedDB): the version is recorded at initialization, and a store refuses to open on a mismatched or missing record, with reset guidance.
- Made ingest fail closed: a Yrs update whose root is not an encoded property identity is refused before the live document mutates; the registration executor refuses reserved collection prefixes exactly like the descriptor-ingest path; resolution refuses `ORDER BY` on JSON subpaths and any subpath on `id`.
- Lifecycle fixes from the review: schema descriptors mark a model announced only after a successful transport send (the previous up-front marking could race a concurrent task into shipping schema-less and lose a delta on an ephemeral receiver); catalog definition loads for announcement are all-or-nothing per model; commit-event generation resolves the model id before draining staged operations; lenient reads honor first-backend-entry-wins so a tombstone shadows later backends; and waiting on a live query surfaces its latched error instead of returning Ok for a dead query.
- Hardened exact-schema binding and admission against aliasing, display-name changes, immutable envelope rebinding, and incorrect offline confirmation.
- Hardened catalog warming and live-query readiness against cancellation, silent peers, cold selection updates, stale generations, and hard-reset races. Catalog-owned initial responses are correlated with their current request and remain owned by the current system generation through schema ingestion and application. Streamed items carry source-query IDs, allowing reset-sensitive catalog traffic to be rejected before schema ingestion or persistence.
- Made `hard_reset` drain in-flight durable warming, ephemeral setup and cache reads, and schema registration before deleting storage. Canceled resets retain their drain barrier for the next attempt, concurrent resets serialize, and an incomplete-reset marker remains set after cancellation so `create`, `join_system`, or startup loading must finish the interrupted reset before publishing a root.
- Re-arms exactly one durable catalog listener for the current reset generation after a replacement system root is ready. Deterministic regressions now cover delayed warming, canceled and concurrent reset attempts, stale registration responses, rootless interrupted-reset recovery, bounded post-reset registration, and stale-generation rejection. This completes #334's acceptance coverage.
- Added strict explicit-ID validation, transaction-local LWW read-your-writes, reference-target clearing, and protocol-handshake timeouts/backoff.

## Why a protocol version bump is required

The serialized contract changed in incompatible ways: events and entity states now carry model IDs, state operations use stable property identity, subscription deltas can include catalog descriptors and source-query provenance, schema registration has new request and response bodies, and resolved `ORDER BY` terms carry property identity.

Maintaining both wire formats would require two codecs and two data-plane semantics throughout the connectors, node, reactor, and storage layers. Protocol version 1 (the 0.10.0 wire; versions number releases, and 0.9.x and earlier are version 0) therefore uses exact version matching. Incompatible peers are refused during Presence negotiation, the websocket server bounds pre-establishment silence, and clients back off after a version refusal instead of reconnecting in a tight loop.

For a rolling deployment, upgrade durable nodes before ephemeral clients. Version 0 (0.9.x) and version 1 (0.10.0) peers do not interoperate during the transition.

## Compatibility

- Version 1 peers require an exact version match; pre-versioning (0.9.x or older) peers are refused.
- There is no legacy-state compatibility: the PropertyKey retirement deleted the legacy decode paths inside the unreleased version 1 epoch. A store written by an earlier build now refuses to open via the storage protocol-version record; development databases reset.
- The metadata catalog itself remains name-keyed as the bootstrap base case.
- Persisted Sled indexes whose encoding metadata does not match the new canonical type are rebuilt rather than reused.
- PostgreSQL physical-column type derivation from canonical metadata remains a follow-up in #343. This PR does not claim that work is complete.
- Catalog-owned reset traffic is fenced. Ordinary user relay queries are not yet bound to a system epoch across a generic `hard_reset`, so late old-system user traffic remains a preexisting follow-up in #360.

## Validation

Current final-tree validation (2026-07-24):

```text
PASS - full default-members suite: cargo test
PASS - benchmark targets compile: cargo check --benches
PASS - focused core/core-types and SQLite, PostgreSQL, Sled, and IndexedDB suites
PASS - complete IndexedDB browser suite in headless Chrome
PASS - cargo fmt --all -- --check and git diff --check
PASS - Format Check after installing Taplo with stable Rust; the two prior failures were nightly rustc ICEs before any format check ran
```

Commands used for the final gate:

```text
cargo test
cargo check --benches
cargo test -p ankurah-core-types -p ankurah-core -p ankurah-storage-sled -p ankurah-storage-sqlite -p ankurah-storage-indexeddb-wasm --lib
cargo test -p ankurah-storage-postgres --test protocol_version
cargo test -p ankurah-storage-sqlite --test protocol_version
CARGO_TARGET_WASM32_UNKNOWN_UNKNOWN_RUNNER=wasm-bindgen-test-runner CHROMEDRIVER=<matching-driver> WASM_BINDGEN_TEST_ONLY_WEB=1 cargo test --target wasm32-unknown-unknown
cargo fmt --all -- --check
git diff --check
```

The gate record for the durable-metadata base is preserved in this description's edit history. The folded identity refactors were reviewed by a multi-track correctness review, and the resulting remediation landed one finding per commit, each independently verified before it landed.

Focused regressions cover exact-schema admission, rename stability, offline reassertion, canonical casting and ordering, index upgrades, immutable schema envelopes, catalog readiness and reset fencing, per-item source-query admission, interrupted and concurrent hard reset, durable post-reset rewarm, stale schema-registration responses, LWW read-your-writes, strict IDs, PostgreSQL narrow-literal/bigint binding, websocket negotiation, and deterministic simulation coherence.

The websocket propagation tests now distinguish cache readiness from remote readiness: cached `query_wait` on an ephemeral node may return after local cache initialization, while tests that mutate immediately after the wait use `nocache` so they wait for remote establishment. This does not claim to resolve the broader subscription sequencing work in #147.

## Follow-ups

- Catalog/schema evolution and tooling: #301, #302, #303, #304, #305, #308, #309, #311, #312, #313, #314, #326, #327
- Descriptor and engine-cache follow-ups: #331, #332, #333
- Identity-refactor relations: fixes #374 (wrong rows after rename); #376 covers the naming half separately; #377 (ORDER BY property identity) is done here; #378 collects the deferred consolidation work (shared guard helpers across engines, the Sled durable-key serialization unification, and related cleanups)
- Existing correctness follow-ups kept out of this epoch: #146, #147, #210, #259, #343, #344, #345, #346, #347, #360
- Build and benchmark infrastructure: #211, #349

Closes #85
Closes #175
Closes #289
Closes #294
Closes #310
Closes #334
Closes #364
Closes #365

## Summary by CodeRabbit

* **New Features**
  * Stable, resolved property identities powering predicate and `ORDER BY` resolution via the durable catalog.
  * Versioned presence handshake with explicit incompatible-peer rejections.
  * A durable protocol-version record for storage engines, and schema registration via `RegisterSchema`.

* **Bug Fixes**
  * Parsing: identifiers like `note` are no longer mistaken for unary `NOT`.
  * Hardened selection/order-by checking and absent-property null/false collapsing.
  * Improved peer/server handshake reliability (timeouts, fatal failures, and rollback on failed peer registration).

* **Documentation**
  * Expanded guidance on canonical value types/casting and amendment retraction audits.


dnorman:
Pushed 695b7415, resolving both spec-surface flags from the review section (maintainer direction on #289):

- **`anchored` descriptor bit (plan decision 18).** `PropertyDescriptor` now records whether `#[property(anchor = ...)]` was physically present, so the RFC 5.8 anchor-reuse guard distinguishes a deliberate rename-back (anchored, `anchor == name`: allowed, display name restored) from the attribute-less retired-name collision (still refused). Free now because RegisterSchema is unshipped; post-release this would have been a versioned descriptor change.
- **Provenance-ordered follow-ups.** Working the rename-back emission exposed a pre-existing defect: the executor parented every follow-up event at the genesis, making all metadata writes after the first mutually concurrent -- a chained rename (a -> b -> c) or an `optional` flip was decided by the event-id hash tiebreak, not recency. Follow-ups (property/model display name, `target_model`, membership `optional`) now parent at the entity's current head and are emitted only when the current catalog value differs. Identical concurrent registrations still mint byte-identical events (same parent, same fields), and an unchanged re-registration is a true no-op (zero events). Model display names can now also revert to the collection name.

Mixed-fleet note: display names follow the most recent registration (an old binary re-asserts its compiled name on startup, and a newer one re-asserts back). This is cosmetic and self-limiting -- addressing is unaffected because each node resolves through its own compiled overlay (RFC 5.8).

New tests: rename-back allowed/refused by anchoredness, chained renames win by recency with advancing heads, membership `optional` flips, model display-name revert. Suite: core 211, ankurah-tests 232, derive 14, ankql 66, sqlite green, wasm check clean.

dnorman:
Two more commits from the maintainer's direction on #289:

- **585267ee (spec)**: erratum 1 ratified with the None-vs-empty rationale (yrs has no mergeable "unset" operation, so `Option<String>` requires the null-capable LWW register); the cast-at-read retype direction recorded in tasks.md Phase C (the coercion half of #303).
- **927cb37c**: `Property::VALUE_TYPE` (plan decision 19), resolving erratum 2 by design change. Custom Property types now declare their own normative value_type through an associated const instead of being assumed "string": `#[derive(Property)]` pins "string" to match its JSON-string serialization, and a hand-written impl producing another `Value` variant declares it (compile-time, zero-cost, no per-field annotation -- the derive emits `<Ty as Property>::VALUE_TYPE` into the schema static). The declared type flows into the registration request, the catalog, and the property-id derivation. Invariant documented on the const: it must equal the `Value` variant `into_value` produces, and changing it for a shipped type is a retype. No existing type re-keys.

Also recorded in tasks.md (maintainer nod): the system-transaction refactor of the registration executor -- replace the hand-built follow-up events with the ordinary Entity/Transaction machinery behind an internal protected-collections capability, with the frozen genesis injected rather than generated. Behavior-preserving (the executor is now correct and test-pinned), so it lands as a follow-up rather than growing this PR.

New end-to-end test: a hand-implemented `Stars(i64)` property type declaring "i64" flows through derive, schema, registration, catalog, and derivation -- incidentally the first in-repo proof that a hand-written Property impl works as a derived-model field. Suite: core 211, ankurah-tests 233, derive 15, ankql 66, sqlite green, wasm check clean.

dnorman:
An independent external review (codex) ran against this branch and returned three findings; all three verified as real and are fixed in ccb358dd:

1. **P1, confirmed: `#[model(id = ...)]` split local and durable identity.** The derive and the `cache_compiled` overlay honored the explicit model binding, but `ModelDescriptor` did not carry it, so the executor derived a different model id from the collection -- local resolution and the durable catalog could silently bind different contracts. `ModelDescriptor` now carries `explicit_id` with the same verify-never-mint semantics as properties (absence hard-fails, collection mismatch hard-fails), and properties/memberships in the request derive under a request-local collection-to-model-id map, matching the overlay.

2. **P1, confirmed: bind-time migration could regress a newer write.** Migrating a name residue onto an occupied id key kept the id entry unconditionally ("id entries are authoritative"), but a residue written while the backend was unbound (v2 mode, cold catalog) can causally DESCEND the loaded id entry -- the old rule discarded the newer value when the catalog warmed. `bind_schema_at(binding, head)` now elects: a head-tip entry is causally maximal and wins outright, an uncommitted residue wins, and otherwise the standard deterministic concurrent event-id tiebreak applies. (Full DAG election needs the apply path's layer machinery; the tip check is exact for the dominant collision and never prefers an entry merely for being id-keyed.)

3. **P2, confirmed: ORDER BY bypassed resolution.** Sort keys now resolve fail-closed under the same rules as predicate references (UnknownProperty on unknown keys, `id` passthrough, collection-qualifier normalization), keeping the PathExpr representation since the resolved display name is the sort key until catalog-bound engines (#312).

Codex's checked-and-cleared list independently confirms the frozen identity/genesis surfaces, the provenance-ordered executor, the catalog lifecycle, and the assembly-time binding funnel. New tests cover all three fixes. Suite: core 212, ankurah-tests 235, jwt-auth 88, derive 15, ankql 66, sqlite green, wasm check clean.

dnorman:
**Identity plane redesigned per maintainer review (2026-07-06): hold review.**

Daniel reviewed RFC 5.1's deterministic derivation and rejected it. Durable-allocated ids replace it: the RegisterSchema executor on the durable node becomes a lookup-or-create upsert by current name and allocates a true EntityId (a real ULID) on first sighting, returning the allocated definitions in a new SchemaRegistered response. Grounds, in brief:

1. Derived ids were SHA-256 bytes carried inside the EntityId ULID newtype -- no timestamp, no randomness, a documented type fib.
2. Derivation keyed identity on the anchor, a fossil of the first field name; the entire 5.8 apparatus (permanent anchor attribute, reuse guard, anchored bit, rename-back semantics) existed only to manage that fossil.
3. Derivation over-converges: re-registering a name after an excision silently resurrects the old identity.

What changes on this branch: schema_id derivation, the frozen genesis encoder, relay-side genesis self-certification, the anchor apparatus end to end, and their golden-vector suites are removed. What does not change: the data plane -- LWW v2/0xA2, schema binding, read rules and the sibling gate, resolution, protocol v2 handshake, catalog protection, provenance-ordered follow-ups.

Sequence: an RFC rev 4 amendment goes to specs/ and #289 for ratification first; implementation lands here after sign-off; the PR title and body will be rewritten for the allocation design. Until then this PR describes the superseded design -- hold review.



---

Update (2026-07-10): the durable-allocation rebuild is complete and review resumed. Comments above this one (now collapsed as outdated) describe the superseded derivation design; the rev 4 ratification trail is on #289.

dnorman:
**Follow-up candidate, noted for consideration (not part of this PR): distinguish authorization failures from infrastructure failures on the read path's registration error.**

Today, when first-use registration cannot run, fetch and live queries surface one collapsed error regardless of cause: `UnregisteredCollection` ("collection 'x' is not registered and could not be registered from this node"). Locally the executor distinguishes `RegistrationError::PolicyDenied` (authorization) from `Mutation`/`Retrieval` (infrastructure), but the wire response is a plain error string and the resolve path drops the cause into a debug trace, so callers cannot tell "you may not" from "something broke".

Two possible shapes, smallest first:

1. Thread the underlying cause into the loud error's message. The server already chose what to say in its error response, and the collection-access gate runs before any lookup, so a denied prober still learns nothing it could not already see. Cosmetic, no wire change.
2. A typed error variant over the wire (structured policy-denial vs failure in the RegisterSchema response), so client code can branch on it. Small protocol addition; would want a look at whether it interacts with the existence-oracle closure before committing to it.

Maintainer to decide whether either is wanted; neither blocks this PR.

dnorman:
Tracked as #326 (labeled `metadata`), for consideration outside this PR.

dnorman:
Known non-blocking follow-up from the review, tracked as #327: `update_remote_query` (core/src/node.rs:1079) silently drops a live-query selection update if it lands in the narrow window before the async relay registration completes, leaving the peer on the old selection. Standard usage (await initialized, then update) is safe; branch-introduced by the id-keyed epoch's move of resolution onto the forward path. Fix and regression test scoped in the issue.

dnorman:
Two fixes landed on this branch out of the cross-PR rebase preparation for #318 (the D1 ingest pipeline), at the maintainer's direction:

1. **The reactor duplicated a change's events once per matching query** (red pin 8a68496c, fix b13287ab). A peer's subscription handler holds every query that peer registered, and two client queries over the same predicate are routine. Since the catalog work, `evaluate_changes` in core/src/reactor/subscription_state.rs appended a change's events to the per-entity update item on every per-query visit, so such a peer received every event twice in one item. No test on this branch caught it because the current applier collapses duplicates during topological sorting before applying; #318 replaces that sort with an id-keyed planner whose state-and-event fast path echoes the item's events into client change notifications, which is where the duplicate becomes visible (two inter_node tests fail on the combination of the two branches; both branches pass alone). The fix gates the fold per entity and change, while distinct changes for one entity in a batch still all contribute their events, preserving the genesis-plus-follow-up shape the append exists for.

2. **The taplo format check fails on generated files once the test suite has run** (fixed in 8521cb0b). The trybuild compile-fail fixtures generate a scratch Cargo.toml under target/tests/trybuild, and the bare `taplo fmt --check` scans it. A fresh checkout masks this, which is why CI stays green while any developer worktree fails the check after one suite run. `.taplo.toml` now excludes target trees.

Full validation gate green at the new tip. One further cross-PR interaction is documented at the top of #318's description (schema registration inherits the atomic commit lane that #318 introduces, which invalidates the durable-partial premise of one test here) and awaits a maintainer ruling; no change to this branch for it yet.

dnorman:
CodeRabbit disposition record, part 1 (inline replies are blocked while a review is pending, so dispositions land here).

The ankql/src/selection/sql.rs finding (unquoted display labels produce invalid SQL for reserved words) is declined by maintainer ruling: that rendering is not SQL and never reaches a database. Nothing consumes generate_expr_sql or generate_selection_sql except the Display impls for Selection and Predicate in ankql/src/ast.rs, and each storage engine translates resolved property identities to physical columns in its own SQL builder. Printing the resolved display label unquoted is the readable rendering of what the statement said. The legitimate residue is the module's misleading name, tracked in https://github.com/ankurah/ankurah/issues/395. Thread resolved.

dnorman:
CodeRabbit disposition record, part 2 (inline replies stay blocked while a review is pending; part 1 is the earlier comment covering the ankql sql.rs rendering finding).

All nine remaining findings were verified against the code and against main for pre-existence. Verdicts:

FIXED on the branch (commits ec79aeb6 and c157d565):

1. core/src/reactor.rs, ORDER BY subpath dropped from the key spec: real. The reactor built sort-key specs with sub_path: None while the paired extraction paths carried the subpath, so orderings differing only by subpath produced equal key specs and an order-by change between them read as no change. The keypart now mirrors the extraction path.
2. storage/postgres/src/lib.rs, reserve the metadata table names: real. A user collection literally named ankurah_meta passes sane_name (the reserved _ankurah_ prefix does not cover it) and would collide with the protocol-version record table. Both engines now refuse their own physical table names as collection ids (postgres and sqlite).
3. storage/postgres/src/lib.rs, column-map table creation race: real. The table is engine wide but was created under per-collection advisory locks, so first opens of two different collections could race its CREATE TABLE. It now takes its own fixed-key DDL lock.
4. storage/indexeddb-wasm/src/engine.rs, reset across three transactions: real, low severity. The wipe now runs as one transaction across entities, events, and property_columns, so an interrupted reset cannot leave the id-to-field map alive after the data is gone.

REFUTED:

5. storage/sled/src/index.rs, primary-key keypart handling: the planner never emits a primary-key keypart (the branch commit "storage: the planner speaks identities and never emits a primary-key keypart" strips it, since the primary key is the natural tiebreak of every index scan), and the planner is the only key-spec producer that reaches the sled index manager. A debug tripwire at sled index ingestion would pin the invariant; noted as optional hardening, not a defect.

PRE-EXISTING on main, spilled to issues rather than fixed here:

6. DDL advisory locks are not cancellation-safe: issue #398. The acquire/release pairing predates the branch; the branch added two more sites.
7. The {collection}_event table naming can alias a sibling collection: issue #399. The convention predates the branch in both SQL engines.
8. sled state, materialization, and index writes are not atomic: issue #400. The write ordering predates the branch; the branch re-keyed materialization without changing it.

OPEN, awaiting a maintainer ruling (thread left unresolved):

9. storage/postgres/src/lib.rs, corrupt durable column-map row is skipped with a warning. SQLite documents the same skip as a deliberate availability choice (a corrupt row hides one property rather than failing the collection), but a hidden assignment means a later miss assigns that property a NEW column, silently splitting its data. Failing closed matches the branch's resolution semantics; keeping the skip favors availability. Ruling requested.

dnorman:
CodeRabbit disposition record, part 3: the one open verdict is now ruled and fixed. The maintainer ruled that an unparseable durable column-map row must fail closed, reversing SQLite's documented skip: a hidden assignment would let the property's next cache miss claim a fresh column and silently split its data. Both engines now refuse to load the map on a corrupt row (commit dd1533ae; the PropertyId::Id pin still seeds last so no row can remap the primary key). All CodeRabbit findings on this PR are now dispositioned: five fixed, one declined by ruling with the rename tracked in #395, one refuted by the planner invariant, three pre-existing spilled as #398, #399, and #400.

extensions/jwt-auth/src/agent.rs
```diff
@@ -247,6 +247,7 @@ impl PolicyAgent for JwtAgent {
         id: &proto::EntityId,
         collection: &proto::CollectionId,
         state: &proto::State,
+        resolver: Option<std::sync::Weak<dyn ankurah_core::schema::CatalogResolver>>,
```
dnorman: Why not just have an inject_resolver method for trait PropertyAgent instead?

core/src/reactor.rs
```diff
@@ -553,23 +652,24 @@ mod tests {
     #[async_trait::async_trait]
     impl crate::node::TNodeErased<TestEntity> for MockNode {
         fn unsubscribe_remote_predicate(&self, _query_id: proto::QueryId) {}
-        fn update_remote_query(
+        fn update_query_selection(
```
dnorman, line 655: why does this need to have its args changed?

dnorman:
Closing in favor of a surgical PR series cut from this branch. At 256 files and +32,739/-8,198, line-by-line review of a single PR cannot establish confidence in minimality. The same content lands as a sequence of small PRs: first moved code and strictly additive files (ankurah-core-types, CatalogManager, and similar, inert where their consumers have not landed), then intermediate contract changes including the StorageEngine trait, then the remaining wiring in dependency order. The model-property-metadata branch stays as the reference tree the cuts are taken from. Findings from the reviews recorded here are tracked in #413 through #417 and #419; the first cut PR will link back to this one.

Issue #313, opened 2026-07-06
Registration executor: system-transaction refactor


dnorman:
Replace the hand-built follow-up events in core/src/schema/registration.rs (follow_up/head_or_genesis) with the ordinary Entity/Transaction machinery behind an internal system-transaction capability: permitted on protected catalog collections, committing through the check_event-gated pipeline, with the FROZEN genesis injected (never generated -- byte identity). generate_commit_event head-parents automatically, making the genesis-parenting bug class (fixed in PR #307) structurally impossible. Behavior-preserving; the executor is correct and test-pinned today. Design notes: specs/model-property-metadata/tasks.md.

dnorman:
**Scope amendment (2026-07-06, from the PR #307 review rounds): this refactor also owns the transactional-commit upgrade.**

The maintainer ruled in session (recorded in `specs/model-property-metadata/rfc.md` section 5.7 and in decision 26 of the `plan.md` index) that registration need not be atomic: the executor commits through the ordinary per-event pipeline, so an agent that allows the resolved plan but denies a constituent event at `check_event` aborts the remainder and leaves the earlier catalog events durable. PR #307 hardened the allocator against the identity consequence -- every lookup miss double-checks durable storage before minting, so a retry converges on the stored ids instead of re-minting (`registration.rs` `*_lookup_checked`, pinned by `partial_registration_retry_does_not_double_allocate`).

What remains for this refactor beyond the mechanical system-transaction move: validate every event of the registration batch first and persist atomically, upgrading "partials are identity-safe" to "partials cannot happen". The "behavior-preserving" framing above still holds for the happy path; the failure path improves.

dnorman:
**Ruling update (maintainer, 2026-07-07): the denial-atomicity half of this upgrade arrives with PR #318; ratified.**

PR #318 (the unified ingest pipeline) reworks commit_remote_transaction to plan and policy-check every event of a transaction before anything commits, and execute_schema_registration (core/src/schema/registration.rs) commits through exactly that lane. Once #318 merges, a denied registration leaves nothing durable. The maintainer ratified this as the new contract, superseding the 2026-07-06 ruling recorded above (registration need not be atomic; a mid-batch denial leaves earlier catalog events durable) for the denial case.

[...]

Issue #274, opened 2026-07-04
RFC: validated ingress with a single ValidatedEvent entry point


dnorman:
Phase 2 workstream D item 6. Tracking: #269; scope: specs/concurrency/phase-2.md. Filed at kickoff alongside the history lifecycle, transactional visibility, and snapshot authority RFCs.

## Motivation

"No unvalidated event touches state" is a convention today, not a structure. Validation (structural checks, attestation verification, policy admission) is applied at some ingress arms and skipped at others; #244 (remote event fetching during BFS bypasses validate_received_event, a gap shared with main) is the existence proof for the class. Every new arm re-decides what to check.

## Proposal

1. A single ingress seam per node converts wire payloads into ValidatedEvent: structural validation (content-hash id verification, clock well-formedness and sortedness, size limits per #246), attestation verification, and PolicyAgent admission, in one place.
2. ValidatedEvent is a newtype whose constructor is private to the ingress module. The staging map stores ValidatedEvent; the applier and the #268 planner accept nothing else. The compiler enforces the invariant instead of review, in the same style as the GetEvents vs SuspenseEvents capability split.
3. Every arm (EventOnly, StateAndEvent, EventBridge, fetch responses, subscription updates, BFS-time remote retrieval) feeds the same seam before staging. State snapshots get the analogous ValidatedState treatment at the same boundary.
4. The PolicyAgent surface is designed once, together with #264's commit-time hook: admission (this RFC) and commit-time validation are two call sites of one coherent API.

## Why standalone rather than folded into #268

#268 owns application semantics (planning, gap replay, typed errors, recovery outcomes) and is already large. This RFC owns the admission boundary and its type. They also land at different points in the workstream D ladder (D1 vs D6). The seam is shared: the pipeline consumes exactly what this ingress produces, so the two RFCs reference each other rather than merge.

## Interactions

- #264 commit-time policy hook: one PolicyAgent surface, two hook points.
- C4 threat model: decides which checks are Byzantine-safe vs trusted-peer; that document is the checklist this ingress implements.
- V5 from the July verification (clock sortedness as an unenforced load-bearing invariant): structural validation gets a permanent home here.
- #266 applied-set: cheap redelivery short-circuit before full validation cost.
- #242 and #243: the fork-based policy preview questions get answered once, at this seam, instead of per call site.

## Non-goals

- New attestation formats.
- Policy semantics changes.
- Wire changes.

## Open questions

[...]

PR #318, opened 2026-07-06
Phase 2 D1+D2: unified ingest pipeline (#268), payload generations and applied-set index (#266)

### LINKED ISSUE #243: Creation events in commit_remote_transaction bypass fork-based policy validation
## Summary

In `commit_remote_transaction` (`core/src/node.rs`), creation events are applied directly to the real entity before the policy check, while update events correctly use a fork. If a non-permissive policy agent rejects the event, the entity is left in a poisoned state.

## Details

For creation events (`is_entity_create()` and head is empty):
- The event is applied directly to the real entity
- Then the policy check happens afterward
- If `check_event` returns Err, the entity already has the event applied but it will never be committed to storage

For update events (handled correctly):
- A fork/snapshot is created
- The event is applied to the fork
- Policy is checked against the fork
- Only if policy passes is the event applied to the real entity

### Comparison with commit_local_trx

`commit_local_trx` in `context.rs` handles this correctly by using forks for all events (both creates and updates).

## Fix

Use the same fork pattern for creation events as is used for updates in `commit_remote_transaction`.

## Impact

- No functional impact with PermissiveAgent (current default)
- Critical with real policy agents
- Related to #242

### LINKED ISSUE #266: RFC: indexed causality via persistent per-event generation numbers
## Summary

Add a persistent, derived **generation number** per event (1 + max of parent generations; 1 for genesis) and use it to accelerate causal comparison, meet finding, and layer computation. Today every `apply_event`/`apply_state` runs a fresh backward BFS whose cost is bounded only by the traversal budget; generations turn most comparisons into cheap monotone checks and give the BFS a principled termination and ordering rule. This is the same maturation step git took with commit-graph generation numbers.

## Motivation

- The comparison state machine re-walks history per call. The accumulator's LRU softens storage round-trips within a call, but nothing amortizes across calls, and deep linear histories are the worst case (the budget plus 4x escalation exists precisely because of this).
- Several verdicts need only an order check, not a walk: an event cannot descend another with an equal or higher generation. Quick rejections for StrictAscends/StrictDescends candidates become O(heads).
- BFS frontier processing ordered by descending generation walks the DAG level-correctly without revisits by construction, and meet search gains a sound stop condition: once every frontier entry's generation is below the minimum generation of the current common set, no new meet candidate can appear, so exhaustion-to-genesis is no longer required to conclude `DivergedSince`.
- Layer computation from the meet is generation stratification relative to the meet; the existing generalized topological iterator keeps its role but seeds and advances become index lookups.

## Proposal

1. **Representation.** `generation: u32` stored alongside each event at commit/ingest time. It is derived data: recomputable from parents, never trusted from peers (verify or recompute on ingest; parents are always resolvable at commit time by the staging discipline, and bridge batches are already applied parents-first).
2. **Storage.** A sidecar column/keyspace per storage engine (sled tree, postgres column, IndexedDB store) plus a lazy backfill migration: compute on first touch for pre-existing events, or a one-shot walk per entity.
3. **Comparison integration.** (a) O(heads) precheck: if `max_gen(subject) <= min_gen(comparison heads' parents)`-style bounds rule out descent, skip straight to the diverged path or the full walk as appropriate. (b) Frontier as a max-generation priority queue. (c) Meet stop condition as above. (d) The traversal budget becomes a true anomaly guard rather than a load-bearing bound.
4. **API.** `GetEvents` grows `get_generation(&EventId)` with a default falling back to event fetch, so engines can adopt incrementally.

## Alternatives considered

- **Interval/reachability labeling (GRAIL-style):** stronger queries, but labels are invalidated by concurrent inserts and are overkill for our access pattern.
- **Per-entity ancestry bitmaps or Bloom filters:** attractive for `dag_contains`-heavy paths, but sizing and invalidation are fiddly; generations deliver most of the win with a single u32.
- **Cache whole comparison results:** heads churn too fast for a verdict cache to pay for itself; generations accelerate every comparison including novel head pairs.

## Non-goals

No wire-format changes (generation is never transmitted as authority); no replacement of content-addressed ids; no change to verdict semantics, which stay pinned by the verification tests and the property oracle.

## Risks and sequencing

Depends on the comparison-core RFC landing first so the hooks land in the simplified structure. The property-test oracle extends naturally: generations can be validated against brute-force depth per generated DAG. Migration risk is low because the data is derived and rebuildable.

## References

- Book: `src/concurrency/causal-comparison.md` (budget, frontier mechanics) on the ankurah.org `201-docs` branch
- Prior art: git commit-graph generation numbers (corrected commit date v2)
- Tactical notes: `archive/201-concurrent-updates-specs` factorization notes items 14-17

## Amendments (workstream B synthesis, 2026-07-06)

Adopted from the adversarial lit-review synthesis, specs/concurrency/lit-review/design-deltas.md (lands with PR #275); each item cites its delta there, which carries the evidence and verdict. Where an amendment conflicts with the text above, the amendment governs.

1. (266-A) Generation is the git v1 topological level: 1 + max of parent generations. The corrected-commit-date (v2) reference is dropped: v2's fix anchors on wall clocks, which this design forbids. The v1 looseness under old-parent reattachment is inherited and unfixable without a trusted clock; consumers (including the D3 horizon) must not treat generation distance as a tight bound.
2. (266-B) The applied-set is a set of event ids. Gap detection is true set difference, never a generation-range test: a "generation below g" membership proxy reproduces Mercurial's crossed-linkrev bug class.
3. (266-C) Correctness requirements, not polish: a self-consistency check (stored generation equals 1 + max of parent generations), a delete-and-rebuild recovery path, a runtime-disable flag with pure-walk fallback, and generation as a small saturating u32 with a deep-chain property-test arm.
4. (266-D) The alternatives rejection stands with the corrected rationale: the query shape (pairwise compare plus meet finding) needs parent edges regardless, so a walk is unavoidable and a tighter reachability filter only trims constants. Revisit 2-hop labeling only if a high-volume dag_contains-without-meet path emerges.

### LINKED ISSUE #268: RFC: unified ingest pipeline with planned application, gap replay, and typed errors
## Summary

Consolidate the several event-application loops into one ingest pipeline with a planning phase and uniform execution semantics, and give application errors a typed taxonomy. This absorbs the planned StrictDescends gap-replay defense (the B3 follow-up from the July remediation) as a property of the plan rather than a patch inside `entity.rs`.

## Motivation

Event application currently happens through parallel loops with subtly different semantics:

- `node_applier` EventOnly arm: stage all, apply each, per-item failure containment, phantom eviction, partial-progress notification (added in the July remediation).
- `node_applier` StateAndEvent fallback arm: near-identical loop, but without the containment semantics the EventOnly arm gained.
- `node_applier` EventBridge arm: stage all, topologically sort, apply parents-first.
- `context::commit_local_trx`: fork-apply for policy checks, then two-phase persist.
- The validate-and-stage preamble is duplicated verbatim between two arms (the branch removed the `save_events` helper that used to unify it).

Each divergence is a place for the next V4/V6-class bug. The remediation fixed the instances; the shape that produced them remains. Separately, the entity-level defense for gap-jumps (an event applied as StrictDescends over staged-but-unapplied ancestors silently orphans their operations) was deliberately deferred as B3 because it did not fit anywhere cleanly: it is naturally a planning concern, not an `apply_event` concern.

## Proposal

1. **Plan, then execute.** An `IngestPlan` built from (entity, staged event set): topological order over the staged set, and for each event the intended action given the projected head. Planning is pure over the staged set plus the comparison machinery.
2. **Gap replay by construction (B3).** If the plan encounters an event whose verdict would be StrictDescends across staged-but-unapplied ancestors, those ancestors are already earlier in the plan by topological order; for the residual case (ancestors staged in a previous delivery but never applied), the planner walks parent edges within staging and schedules them first. The gap-jump loss class disappears structurally.
3. **Uniform execution.** One executor owning the invariants that today live by convention at each call site: apply-before-commit per event, commit-before-set_state per entity, per-item failure containment with applied-subset notification, phantom eviction, batched reactor notification. EventOnly, StateAndEvent fallback, and EventBridge become three ways to feed the same pipeline; `commit_local_trx` reuses the planner with its policy-check phase.
4. **Typed errors.** Replace stringly `MutationError::General`/`InvalidUpdate` at the application boundary with a taxonomy distinguishing policy denial, lineage rejection (disjoint, invalid over empty head), storage failure, and transport/validation failure, aggregated per item in the existing `ApplyError::Items` shape. Ack payloads and logs become actionable; callers stop matching on strings.
5. **Recovery hook.** The needs-state recovery deferred from the V6 fix (unknown entity referenced by EventOnly: request a snapshot from the peer) becomes a planner outcome (`NeedsState`) instead of an error, handled by the applier that owns peer communication.

## Non-goals

No wire changes; no changes to comparison semantics; no reordering of the durability invariants (stage before head, commit before state), which the pipeline centralizes rather than alters.

## Sequencing

Independent of the comparison-core and indexed-causality RFCs. The conformance surface is already strong: V4's bridge tests, V6's containment test, C2's atomicity test, and the durable/ephemeral suites pin the behavior the pipeline must preserve. This RFC supersedes filing B3 as a standalone issue.

## References

- Book: `src/concurrency/factorization.md` (current seams and invariants) on the ankurah.org `201-docs` branch
- Remediation history: branch `archive/201-concurrent-updates-specs` (V4, V6, V7 in the verification review; factorization notes items 10-13, 22-23)

## Amendments (workstream B synthesis, 2026-07-06)

Adopted from specs/concurrency/lit-review/design-deltas.md (lands with PR #275). Where an amendment conflicts with the text above, the amendment governs.

1. (268-A) Item 3 gains an explicit idempotency requirement: re-applying an already-committed event is a no-op, including across an ack-loss retry. C6 gains an arm that duplicates the ack and asserts no double apply.
2. (268-B) Disjoint is reachable only through the grounded exploration boundary, never via BudgetExceeded. A dangling-parent event stays staged (buffered) rather than becoming a hard drop.

Status: both amendments are incorporated in the D1 implementation plan and its in-flight implementation (PR #318); the staging area and typed NeedsEvents outcome are the buffering mechanism.

dnorman:
## Summary

This PR collapses Phase 2 workstreams D1 and D2 into one draft for review.

It builds one ingest pipeline for every event application path, then adds generation-aware event identity, per-tip generation metadata, bounded staging, applied-event tracking, and consistent persistence across the native and browser storage engines.

The collapsed stack was rebased onto `model-property-metadata` at `4fdc2ad1`. The latest correctness hardening is commit `33369d98`, followed by a three-commit review follow-up batch.

Closes #268. Closes #243. Closes #266. Tracking issue: #269.

## What changed

- Unified planning and execution for remote updates, event bridges, state snapshots, fetch responses, system join, and local and remote commits.
- Added bounded node-held staging with typed retry behavior and descendant re-drive when missing history arrives.
- Made remote transactions policy-check every entity before phase two and reject mixed-model groups.
- Added a mandatory `generation: u32` to event content and sealed it into `EventId`.
- Added `GClock` per-tip generation metadata to wire state, resident state, PostgreSQL, SQLite, Sled, and IndexedDB.
- Added a bounded in-memory applied set for constant-time redelivery checks.
- Restored safe persistence elision with reset-epoch and canonical-resident checks.
- Added protocol v4 compatibility checks and bumped the workspace crates to `0.10.0`.

## Correctness hardening in `33369d98`

- One reset fence now covers each apply, commit, persist, and notification span. Transactions and network work admitted before a reset cannot leak into the successor system.
- Remote lineage fetches validate requested ID, entity, and model before caching anything.
- Failed or stale plans clean up only their own staged batch, while retryable missing-history items remain buffered.
- State cargo is identity-scoped, committed parent-first, and re-driven without dropping notifications.
- Peer teardown and hard reset drain pending requests and update acknowledgements. Descriptor acknowledgements are bound to the original epoch and live connection.
- Schema registration carries its originating epoch through lookup, durable commit, readiness, and catalog publication.
- PostgreSQL, SQLite, Sled, and IndexedDB reject stored event payloads that do not recompute to the stored event ID. Generation decoding is range checked.
- Descriptor-dependent messages are not sent when a complete schema bundle cannot be assembled.

## Compatibility

This is an intentional wire and disk break:

- Peers must use protocol v4 exactly.
- A fresh database is required. There is no migration for pre-Phase-2 event or entity rows.
- The workspace version is `0.10.0`.

## Validation

Completed at `33369d98`:

- GitHub CI, all five checks, including the full native test job and the browser WASM suite.
- `RUST_BACKTRACE=1 cargo test`, `cargo fmt --all -- --check`, `taplo fmt --check`, prettier over the react example.
- The criterion bench smoke under the repository's memory watchdog (`scripts/memwatch.sh 8192`).
- Expanded oracle runs: the 5000-seed comparison-versus-reachability property oracle and the generation-corruption immunity matrix at 2000 randomized, 200 large, and 100 wide seeds.

Still pending: the separate `tests-wasm` Playwright suite (never part of this repository's standing gate; worth one run before the ready flip).

## Review status

This PR remains draft. The final detailed branch review is COMPLETE at `33369d98`: merge seams verified clean against both parents, all 21 findings of the pre-remediation review independently verified fixed, and the program's invariants confirmed intact (generation sealed in event identity with model attribution excluded; generation values never reach comparison verdicts; the persist funnel refuses non-canonical residents). The review's follow-up batch rides on top: a release-mode epoch guard on fenced response delivery, a reconnect backoff reset after established connections, and re-pinned staging-flood assertions. Remaining liveness follow-ups are tracked in #366 (staging backpressure wedge and unbounded update retry tasks, with #274 as the intended bound).

The maintainer's read and the ready flip are the remaining gates.

## Full record

Detailed history, in reading order:

- #328: the D2 ladder record (milestones M1 to M6, red-test inventories, oracle arms, close-out evidence), resolved as merged into this branch when the stack was collapsed.
- #269 milestone comments: D2 M1 4910463394, M2 4929479575, M3 4930645769, M4 4932828930 plus remediation addendum 4937098423, M5 4942109007, M6 4948587186, and the post-ladder remediation and collapse record 4949282688.
- #266: the D2 RFC; comment 4909916652 records the deferred protocol-version batch discharged by this branch (protocol v4, exact-match compatibility, workspace 0.10.0).
- #313 comment 4907664087: the registration denial-atomicity ratification this branch implements.
- #366: the staging backpressure liveness follow-ups from the final review.
- #335: the generation tripwire escalation discussion (soft posture confirmed; hardening question parked there).


dnorman:
D1 is ready for review: M1 through M8 landed red-first, R1 through R10 green, full gate green at the tip, and the ready-flip evidence block is at the bottom of the description (500-seed swarm with a pre-existing #321-class control finding, bench comparison against both baselines with the bridge_catchup pin holding at 39.0 us/event, and the quantified no-op-delta write cost with its E2 elision candidate).

A four-dimension review panel ran over the whole diff before this flip; its one correctness major (retention sweep vs acked carryover orphans under transient failure) is fixed red-first in the final three commits, along with the Get lane's PerItem containment and the doc/spec reconciliation of the two in-flight amendments. A codex cross-model pass is running separately and any findings from it will follow as review comments.

Suggested review order: commit-by-commit follows the milestone ladder; the two amendment sites (node.rs plan_and_check_entity_group, entity.rs StrictDescends arm) carry the maintainer rulings in place.

dnorman:
Codex round triaged; the PR was briefly re-drafted while its findings landed and is ready again.

Verdicts on the four codex findings, each verified against the code before acting:

1. Commit-failure persist (BLOCKER, confirmed): fixed red-first in 4fd5b9ca/e68288df. The executor now suppresses the plan's persist when an applied event's commit failed (the buffer lags the log's coverage, the direction gap replay already repairs), keeps that event's body staged, and the planner schedules head-contained-but-unstored members so the integrated-but-unstored backfill can commit them on redelivery. Unit pins cover the suppression, the retention, and the full recovery loop.
2. Attestation drop (major, confirmed): StagingArea::restage replaces the staged envelope in place and the shared phase one uses it; an attesting-agent pin asserts the attestation survives to the committed event. This was a regression against the pre-D1 lanes on BOTH commit paths.
3. Comment staleness (major, partially adopted): the state_apply module doc drift was real and is fixed. The milestone/claim-id shorthand stays: it is the repo's traceability convention, and the spec references are load-bearing.
4. Arm dedup (minor): already collapsed to ExecutionOutcome::needs_state_error in the panel round; codex reviewed the pre-refresh packet.

The mid-round CI failure (sim_coherence read_your_writes) reproduced 0/10 locally at the same commit and is the pre-existing #321 wall-clock class documented there.

dnorman:
Re-drafted at the maintainer's request pending the rebase onto main after #307 merges. The rebase is prepared: the stack replays with three trivial conflicts, resolutions are recorded with git rerere, and a preview of the rebased stack plus the three cross-PR fix commits is on branch phase-2-d1-ingest-on-307 with the full validation gate green at its tip. The three interactions (duplicate events in subscription updates, registration inheriting commit atomicity, taplo vs generated trybuild TOML) are summarized at the top of the description. The current commits remain the review surface; their content does not change under the rebase.

dnorman:
Rebase close-out, 2026-07-11

- Collapsed D1 and D2 stack rebased onto `model-property-metadata` at `4fdc2ad1`.
- Replayed the full Phase 2 ladder and adapted model envelopes, property ids, catalog follow-ups, raw-engine resolvers, simulator metadata, and GClock round trips to the metadata contract.
- Added the receiver-side duplicate-fragment regression red-first, then collapsed untrusted wire duplicates by event id before staging and change publication.
- Recorded the ratified registration atomicity contract from #313 comment 4907664087 and pinned zero durable catalog rows after denial plus exact retry identities.
- Added the pre-D2 peer rejection regression red-first, bumped `PROTOCOL_VERSION` from 3 to 4, documented the wire and on-disk break plus fresh-database requirement, and bumped all 24 workspace packages to `0.10.0`.
- Preserved `phase-2-d2-generations` at `36157754`; removed the spent preview worktree and local and remote preview refs.

Local acceptance is green: the full native three-package suite, plain native `cargo check`, isolated core and IndexedDB Wasm checks, Chrome IndexedDB tests, benchmark compilation, Rust and Taplo formatting, and the 8 GiB memwatch Event DAG smoke with all 26 cases.

GitHub CI is green across tests, README verification, format, Wasm browser tests, and Event DAG benchmark smoke. PR #318 remains in draft for the maintainer review flip.

PR #320, opened 2026-07-06
Phase 2 E-shapes: event DAG shape scale tier and lane performance baseline

dnorman:
Workstream E-shapes of concurrency phase 2: event-DAG-shape scale and per-lane end-to-end performance testing. Two measurement instruments, clearly separated, plus the spec amendment that schedules them. Measurement only: no core/src changes of any kind. Tracking #269.

Revision note: the initial version of this PR mislabeled a fresh-fetch benchmark as bridge catch-up; that is corrected in `0f79124b` (details under Perf tier below), with the wrong-lane numbers replaced and a permanent lane guard added so the mistake cannot silently recur.

### 1. Event-DAG-shape scale tier (deterministic correctness and memory at scale)

`tests/tests/sim_event_dag_shapes.rs`, driving the C1 simulation harness at env-scaled event-DAG-shape scale. Four shapes, each asserting the C1 convergence family (all-node convergence, no lost write, no phantom, head antichain validity) at scale plus peak heap use:

- `multi_entity_churn`: N entities across all origins, settle, then edit each from a different origin (extends the C1 churn shape upward).
- `deep_single_entity_history`: one entity with a chain of SHAPE_DEPTH edits; every node catches up through the scheduler. (Deterministic view; catch-up WALL TIME lives in the perf tier.)
- `wide_concurrent_antichain`: many concurrent siblings off one shared fork, forming a genuinely wide head (validated non-trivially by the antichain invariant, `max_head_len` reports the width), then merged to one tip.
- `subscription_fanout`: SHAPE_QUERIES overlapping live queries held open while entity churn runs.

Memory is measured by a test-only counting global allocator (`tests/src/sim/alloc.rs`, a small `GlobalAlloc` wrapper tracking live and peak bytes), installed only in this one test binary. A process-wide measurement lock serializes the measured regions so the process-global allocator peak stays attributable under parallel `cargo test` threads. Bounds are LOOSE order-of-magnitude catastrophe guards (an unbounded buffer, an accumulator that never releases, a per-event leak), not thresholds.

Scale knobs (defaults small enough for a normal `cargo test`; nightly variants scale up):

- `SHAPE_ENTITIES` (200), `SHAPE_DEPTH` (200, the deep-history chain depth), `SHAPE_ANTICHAIN` (64), `SHAPE_QUERIES` (16), `SHAPE_SEEDS` (3).

Memory-attribution caveat (design-deltas E-A / 271-D): the bounded-by-divergence-window target governs the reverse walk; a sequence-CRDT tombstone set is instead lower-bounded by deletion history and folded by sealing, not by the streaming consumer. The scenarios do NO deletions, so their peaks never misattribute that floor; the doc states that a future deletion scenario must set its bound from the deletion count.

### 2. Perf tier (real wall-clock, advisory)

`tests/benches/lane_perf.rs`, a criterion per-lane end-to-end suite (`harness = false`) over REAL multi-threaded tokio and the production `LocalProcessConnection` connector (NOT the sim transport). Each case uses `iter_custom` so per-iteration setup (nodes, seeded history, connections, subscription) is excluded from the measured region. NOT run in the normal CI test job.

Delta-lane discipline (the `0f79124b` correction): a fresh client's fetch sends empty known_matches and is answered by a full StateSnapshot (`generate_entity_delta` Case 3), never an EventBridge. The two lanes are now measured separately, each with a permanent lane guard derived from the arms' observable difference (the bridge arm commits each event to the client's local storage; the snapshot arm commits none):

- `fresh_fetch_snapshot`: fresh-client first fetch at depth-N server history. Depth affects only excluded setup; the measured region is snapshot adoption and is roughly flat in depth. Lane guard: zero locally committed events.
- `bridge_catchup` (TRUE EventBridge lane, stale-client shape): the client fetches once so it persists the entity at a soon-stale head, disconnects, the server advances by GAP edits, then the measured region is reconnect + re-fetch, where known_matches carries the stale head, the server builds the bridge (Case 2), and the client applies GAP events through the bridge arm. Guards: final state reached AND at least GAP locally committed events, so a snapshot fallback fails the bench rather than recording the wrong lane.

Lane verification evidence: a one-off instrumented run of the exact shape printed `stored_events=0` after the fresh fetch (snapshot lane) and `stored_events=25 gap=25` after the stale re-fetch (bridge lane); the instrumentation was removed and the stored-events assertions remain as permanent guards in both groups.

Sim-harness timings are never reported as performance numbers (single-threaded virtual transport); wall-clock comes only from this tier.

## Methodology and numbers

Perf-tier medians captured on this machine (Apple M4 Max, 16 cores, 128 GB, macOS 26.5.1 build 25F80, rustc 1.97.0-nightly ca9a134e0, AC power; same machine as the E1 micro baseline), all groups in one run. Full disclosure and reading notes in `specs/concurrency/LANE-BASELINE.md`. Command: `cargo bench -p ankurah-tests --bench lane_perf`.

| Group / case                                    | Param | Median    | Derived          |
| ----------------------------------------------- | ----- | --------- | ---------------- |
| single_writer_commit / durable_lww_overwrite    | n/a   | 37.06 us  | ~27.0k commits/s |
| commit_to_subscriber_latency / server_to_client | n/a   | 84.37 us  |                  |
| fresh_fetch_snapshot                            | 100   | 110.06 us |                  |
| fresh_fetch_snapshot                            | 1000  | 132.61 us |                  |
| fresh_fetch_snapshot                            | 5000  | 136.98 us |                  |
| bridge_catchup                                  | 100   | 3.325 ms  | ~33 us/event     |
| bridge_catchup                                  | 1000  | 38.41 ms  | ~38 us/event     |
| bridge_catchup                                  | 5000  | 196.2 ms  | ~39 us/event     |
| subscription_establishment                      | 10    | 379.87 us |                  |
| subscription_establishment                      | 100   | 3.412 ms  |                  |
| subscription_establishment                      | 1000  | 39.24 ms  |                  |

Reading notes: `bridge_catchup` scales linearly with gap depth at ~33-39 us per bridge-applied event (~25-30k events/s), the same order as a local commit, which is coherent (each bridge event is validated, staged, applied, and committed on the client, and the server pays the causal comparison plus backward walk). `fresh_fetch_snapshot` is roughly flat in depth, as its lane predicts. `subscription_establishment` grows roughly linearly in resident-entity count. Full caveats in the baseline doc.

Event-DAG-shape tier peak memory at default scale (peak-above-baseline, from the `SHAPEMEM-SUMMARY` log lines), all well under their loose ceilings:

| Scenario                    | Scale         | Worst peak-above-baseline | Ceiling |
| --------------------------- | ------------- | ------------------------- | ------- |
| multi_entity_churn          | 200 entities  | ~30 MiB                   | ~228 MiB |
| deep_single_entity_history  | 200 depth     | ~43 MiB (first-run seed 0)| ~228 MiB |
| wide_concurrent_antichain   | 64 writers    | ~21 MiB (head width 64)   | ~192 MiB |
| subscription_fanout         | 16 queries    | ~22 MiB                   | ~161 MiB |

## Spec amendment rationale

`specs/concurrency/phase-2.md` workstream E: the event-DAG-shape scale and per-lane harness is split out of E as an explicitly-scheduled item that lands DURING workstream D, because it is not an optimization. It is the instrument that guards the D refactors as they land and feeds E2 with the before-numbers, so it must exist before D, not after. E2 (optimization, after D) is left unchanged. The E-shapes checklist line is ticked; no other checklist lines are touched.

## Nightly wiring

`.github/workflows/nightly-oracle.yml` gains a third job (`event-dag-shapes-nightly`) running the event-DAG-shape scale tier at env-scaled scale, with matching `workflow_dispatch` inputs. NOTE: this PR touches a workflow file, so per token scope it must be merged by the maintainer personally.

## Gate evidence

Full validation gate, all green on this branch (re-run after the `0f79124b` correction):

- `cargo test -p ankurah-tests`: all binaries pass, 0 failed (sim_event_dag_shapes: 4 passed, 4 ignored, ~15s, inside the smoke budget).
- `cargo test -p ankurah-core`: 0 failed.
- `cargo test -p ankurah-jwt-auth`: 0 failed.
- `cargo check -p ankurah-core --features wasm`: finished, 0 errors.
- `cargo fmt --all -- --check`: clean.
- `taplo fmt --check`: clean.
- `grep -rP "[\x{2013}\x{2014}]"` over all changed files: no matches (no em or en dashes).

## Constraints honored

- Measurement only, no optimization, no core/src changes of any kind.
- Perf tier not in the normal CI test job; the event-DAG-shape smoke adds ~15s (< ~30s budget).
- Wall-clock only from the perf tier; sim timings never reported as performance.

## Nomenclature

Renamed from E-vol on 2026-07-06 at maintainer request. This tier scales event DAG shape under the deterministic sim (deep single-entity histories, wide concurrent antichains, multi-entity churn, subscription fan-out); it is not volume testing. Volume testing proper (millions of rows, thousands of concurrent ephemeral nodes against a durable node) is tracked separately in #324 and #325. `macro_perf.rs` is now `lane_perf.rs` and `MACRO-BASELINE.md` is now `LANE-BASELINE.md` because "macro" collides with Rust macros and these are per-lane end-to-end benchmarks.

### DIFF
--- a/.github/workflows/nightly-oracle.yml
+++ b/.github/workflows/nightly-oracle.yml
@@ -37,6 +37,26 @@ on:
         description: "First dag_seed value to run (ORACLE_SEED_BASE)"
         required: false
         default: "1"
+      vol_entities:
+        description: "Volume tier: entities in the N-entity churn shape (VOL_ENTITIES)"
+        required: false
+        default: "5000"
+      vol_events:
+        description: "Volume tier: chain depth for the deep-history shape (VOL_EVENTS)"
+        required: false
+        default: "5000"
+      vol_antichain:
+        description: "Volume tier: concurrent writers in the wide-antichain shape (VOL_ANTICHAIN)"
+        required: false
+        default: "512"
+      vol_queries:
+        description: "Volume tier: live queries in the subscription fan-out shape (VOL_QUERIES)"
+        required: false
+        default: "128"
+      vol_seeds:
+        description: "Volume tier: seeds per scenario (VOL_SEEDS)"
+        required: false
+        default: "20"
 
 jobs:
   oracle-scale-out:
@@ -103,5 +123,48 @@ jobs:
         with:
           name: sim-swarm-failure-output
           path: sim-swarm-output.log
+          if-no-files-found: warn
+          retention-days: 30
+
+  # The E-vol volume tier at nightly scale: the same four volume scenarios the
+  # normal test job runs at small defaults (N-entity churn, deep single-entity
+  # history, wide concurrent antichain, subscription fan-out), driven up via the
+  # VOL_* env knobs and the `--ignored` nightly variants. Each scenario asserts
+  # the C1 convergence invariants at volume and a loose order-of-magnitude peak
+  # memory bound (counting global allocator). A VOLMEM bound breach or an
+  # invariant violation fails the job; the VOLMEM lines in the log are the
+  # per-scenario peak-memory record. This is the deterministic volume instrument
+  # only; wall-clock perf numbers come from the criterion macro suite, which is
+  # advisory and NOT run here.
+  volume-tier-nightly:
+    name: Volume tier (env-scaled)
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Set up Rust
+        uses: dtolnay/rust-toolchain@stable
+
+      - name: Cache dependencies
+        uses: Swatinem/rust-cache@v2
+
+      - name: Run the volume tier at scale
+        run: |
+          set -o pipefail
+          VOL_ENTITIES="${{ github.event.inputs.vol_entities || '5000' }}" \
+          VOL_EVENTS="${{ github.event.inputs.vol_events || '5000' }}" \
+          VOL_ANTICHAIN="${{ github.event.inputs.vol_antichain || '512' }}" \
+          VOL_QUERIES="${{ github.event.inputs.vol_queries || '128' }}" \
+          VOL_SEEDS="${{ github.event.inputs.vol_seeds || '20' }}" \
+            cargo test -p ankurah-tests --test sim_volume -- --ignored --nocapture 2>&1 | tee volume-output.log
+        env:
+          RUST_BACKTRACE: 1
+
+      - name: Upload failure output
+        if: failure()
+        uses: actions/upload-artifact@v4
+        with:
+          name: volume-tier-failure-output
+          path: volume-output.log
           if-no-files-found: warn
           retention-days: 30
--- a/.github/workflows/nightly-oracle.yml
+++ b/.github/workflows/nightly-oracle.yml
@@ -37,24 +37,24 @@ on:
         description: "First dag_seed value to run (ORACLE_SEED_BASE)"
         required: false
         default: "1"
-      vol_entities:
-        description: "Volume tier: entities in the N-entity churn shape (VOL_ENTITIES)"
+      shape_entities:
+        description: "Event-DAG-shape tier: entities in the N-entity churn shape (SHAPE_ENTITIES)"
         required: false
         default: "5000"
-      vol_events:
-        description: "Volume tier: chain depth for the deep-history shape (VOL_EVENTS)"
+      shape_depth:
+        description: "Event-DAG-shape tier: chain depth for the deep-history shape (SHAPE_DEPTH)"
         required: false
         default: "5000"
-      vol_antichain:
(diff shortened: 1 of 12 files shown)


dnorman:
Scope clarification from the maintainer (2026-07-06), recorded so this baseline is not mistaken for the larger program: volume testing in the maintainer's intent means TRAFFIC and DATA volume (millions of rows, thousands of clients), now filed as #324; perf testing means load and query latency over realistic datasets (100k rows, tiered suites over time), now filed as #325. This PR's tiers (in-process sim volume, sync-machinery macro baseline) are the first instruments, not the whole program.

Issue #271, opened 2026-07-04
RFC: history lifecycle with sealed-prefix checkpoints and a concurrency rejection horizon


dnorman:
Phase 2 workstream D item 3. Tracking: #269; scope: specs/concurrency/phase-2.md. Filed at kickoff alongside the transactional visibility, snapshot authority, and validated ingress RFCs.

## Motivation

Event history is append-only and unbounded today. Nothing compacts, prunes, or forgets; the compaction and GDPR story is flagged as "Discuss" in DESIGN_GOALS. Two costs grow with history depth:

1. Storage: every event since genesis is retained forever on durable nodes.
2. Traversal: causal comparison walks backward with the retrieval budget as its only backstop. The budget is error-shaped, not policy-shaped. A months-stale branch arriving today cannot be rejected as too old; it either forces a deep re-layering merge (thrash) or dies with BudgetExceeded, which is an anomaly report rather than a decision.

Entity identity is anchored at the creation event (Disjoint detection walks to genesis), so naive pruning would break identity and disjointness checking. Sealing has to preserve that anchor.

## Proposal

1. Sealed-prefix checkpoints. Choose a seal clock C (an antichain of the entity's DAG). Fold all history at or below C into an attested snapshot: per-backend state buffers plus the seal clock and provenance metadata. Prune events strictly below C. Comparison and layer traversals clamp at C; the seal boundary behaves like a pseudo-genesis for walks.
2. Genesis attestation. The checkpoint carries an attestation binding the seal to the entity's creation event id, so Disjoint detection and entity identity survive pruning. A peer retaining full history can verify the binding; peers that sealed compare identity through the attested genesis id instead of walking to it.
3. Concurrency rejection horizon. An event whose parents fall below the seal cannot integrate normally. That failure becomes an explicit policy decision instead of an accident: reject as too old, or route to a designated rebase or merge path. The decision surfaces through PolicyAgent so applications choose their horizon. Never silently drop.
4. Soft horizon via generations. Once #266 lands, staleness distance is deterministic: gen(head) minus gen(meet). A policy can use generation distance as a soft horizon well before the structural seal is reached.

## Hard requirements

- The accept or reject decision must be deterministic and convergent across nodes. A policy where one node rejects what another accepts is a permanent partition machine.
- Every candidate policy needs a DoS analysis: forged parentage that forces deep merges or deep walks must not let an adversary spend our resources unboundedly.
[...]

PR #328, opened 2026-07-08
Phase 2 D2: payload generations plus applied-set index (#266)

### LINKED ISSUE #266: RFC: indexed causality via persistent per-event generation numbers

dnorman:
> **STACKED ON #318, DO NOT MERGE BEFORE IT.** This branch is based on
> `phase-2-d1-ingest` (PR #318, currently a draft pending its own rebase).
> The base of this PR is that branch, not `main`; it must land after #318.
> When #318 rebases, this branch re-parents onto the rebased tip and
> force-pushes with `--force-with-lease`.

Tracking #269. Closes #266.

Implements D2: a mandatory per-event `generation` number (git-v1 topological
level) folded into the event identity, plus the applied-set index and the
comparison accelerations it enables. Plan of record: REV 5, the amended
plan the maintainer approved 2026-07-09 (it governs on conflict with the
earlier revisions). The ladder is sequential (M1 to M6); this PR opens at M1 per the
repo's draft-at-first-milestone convention.

## Fresh database required (release note)

`generation: u32` is a mandatory field inside every event's hashed content:
there is no way to represent an event without one, and there is one id
formula for all events. This is a wire-format and on-disk break, expressly
approved by the maintainer without a compatibility requirement (#266
comments 4908897445 and 4909788267). D2 ships with a minor version bump and
**requires a fresh database**: a pre-D2 database is invalid and its engines
fail loudly on the new event shape. A migration script is deliberately
deferred (maybe-later, out of D2 scope). As of M4 the persisted entity
STATE record changes shape too: postgres and sqlite entity tables gain a
NOT NULL per-tip generations column parallel to the head, and indexeddb
entity records gain the matching key, so pre-M4 state rows also fail
loudly at read instead of reconstituting silently (postgres and sqlite
length- and range-checked against the head; indexeddb fails on the
missing key and, since the M4 remediation and the M6 close-out, rejects
any stored generation that is not exactly a u32 on BOTH of its
generation read paths, the entity-record GClock decode and the event
column; its fetch_states path still silently drops undecodable ROWS, a
pre-existing pattern now recorded as known gap G-11 in
specs/concurrency/threat-model.md rather than redesigned here). Same
ruling, same remedy: fresh database.

## Milestone status

- [x] **M1 event shape epoch.** `generation: u32` on `proto::Event` and
  `EventFragment`, inside the `EventId` hash; the From impls, receive path,
  and all four storage engines carry it (postgres/sqlite NOT NULL column,
  indexeddb key, sled transparent). `has_event` lands on `StorageCollection`
  with cheap per-engine overrides. Counting `GetEvents` wrapper in
  tests/common. Reds R-D2-1a/1b green.
- [x] **M2 stamp plus verify.** The commit lane stamps
  `1 + max(parent generations)` read from the parent EVENT payloads (genesis
  exactly 1; on ephemeral nodes a snapshot-adopted head's parents resolve
  from a durable peer, a pre-accepted per-commit cost), and every
  parents-local admission boundary verifies the equation: the executor loop
  checks before each apply, commit-lane phase one denies a whole transaction
  before anything durable, and a mismatch is the new typed
  `LineageRejection::GenerationMismatch`. Admissions whose parents are not
  locally in hand (adopted-history: the integrated-but-unstored backfill,
  state-adoption cargo, adopted-head extensions) admit WITHOUT verification
  and record their ids in a bounded in-memory unverified set (ids only, FIFO
  at 65536), the M5 eligibility input. The generation registries the M1
  review banned are deleted (core/src/test_gen.rs, tests/src/gen.rs, the sim
  Workload map): every test stamp now reads parent payloads or states an
  explicit adversarial claim. Reds R-D2-2a (stamps equal brute-force depth
  on linear/diamond/crossed), R-D2-2b (mis-stamp rejected typed on both
  boundaries, genesis rule, correctly stamped twin applies), and R-D2-2c
  (adopted-backfill admission lands in the unverified set) verified failing
  at their commits, green at the tip; plus an indexeddb browser round-trip
  of the generation including the u32::MAX sentinel (review F2 ruling), run
  headless in Chrome. Full local gate green at every boundary.
- [x] **M3 applied-set plus fast paths.** Each resident entity carries a
  bounded in-memory applied-set (event ids only, FIFO at 65536,
  constructor-configurable, lazily allocated, never forked, never
  persisted, never shared) recording exactly what a COMPLETED persist of
  that resident's state buffer proves covered: rows are inserted only
  after a `set_state` returns Ok, on all three persist lanes (the ingest
  executor, the shared state-apply, and the local commit lane, whose
  verbatim duplicate of `save_state` now routes through the shared
  `NodePersist` machinery), and the commit-failure suppress-persist path
  inserts nothing. Consumption is a single site: `apply_event` re-reads
  membership under the head lock immediately before the comparison walk
  would run, so an applied-and-persisted redelivery is a zero-fetch O(1)
  no-op; the planner keeps its head-containment preresolution and its
  backfill escape untouched. Also in this milestone, per the plan
  amendment approved 2026-07-09: the StateAndEvent cargo ruling reverses
  one M2 arm, so cargo generations are checked BEFORE the state applies
  and a provably mis-stamped cargo event aborts the whole update item
  with the typed lineage error (no adoption, nothing stored), while
  unverifiable cargo still stores and records acceleration-ineligible;
  and the event DAG criterion bench, broken since M1 because benches sit
  outside the gate's compile surface, is repaired with honest
  parents-in-hand stamping and a benches compile check joins the gate.
  Reds R-D2-3a (redelivered persisted event served by the O(1) skip with
  zero event fetches), R-D2-3b (an event durably committed while its
  plan's persist was suppressed is NOT recorded as incorporated; recovery
  records it only after the backfill commit and a completed persist),
  R-D2-3c (cap eviction changes cost, never outcome), and the section L
  cargo-abort red all verified failing at their exact commits with the
  observed output in the commit bodies, green at the tip. The memory
  audit tier (sim_event_dag_shapes) peaks are within noise of the
  M2-era values.
- [x] **M4 head-generation materialization (GClock), persist marker, and
  reset epoch.** The materialized head now carries per-tip generations via
  the new `proto::GClock` (`Vec<(u32, EventId)>`; canonical entry order
  pinned, generation-major with EventId tiebreak, deduplicated and
  normalized on every construction path including wire deserialization;
  serde determinism unit-pinned). Exactly three homes: `proto::State` (and
  therefore the wire StateFragment), the resident's in-memory entity
  state, and each engine's entity record (postgres NOT NULL bigint[]
  column parallel to the head array, sqlite NOT NULL TEXT column holding a
  JSON u32 array, indexeddb entity-record key, sled transparently in the
  bincoded struct), so rehydration reconstitutes it without reading
  events; `proto::Clock` and `Event.parent` are untouched (no id-formula
  change). Commit stamping now derives from the resident's materialized
  head generations under the same read lock as the head: synchronous, no
  event retrieval on any node (the M2 ephemeral peer fetch at commit is
  gone). Admission verification resolves parents covered by the
  materialized tips (head-parented commits, in-order bridge steps,
  subset-parented arrivals on multi-tip heads) from the materialization
  with zero payload reads and falls back to the M2 local payload reads
  only for uncovered arrivals; a durable node never adopts a wire-carried
  annotation uninspected (a structural mismatch with the head or a
  contradiction with a locally held tip event rejects typed, the latter
  as `LineageRejection::GenerationMismatch`, the former as the new
  `HeadGenerationsMismatch`), while ephemeral nodes adopt inside the
  state's trust envelope. The persist-currency marker (reset epoch plus
  head) lands on the resident, stamped only by the shared persist funnel
  (`NodePersist`, all three lanes) after a completed `set_state`, with
  the node's memory-only reset epoch captured BEFORE the persist begins;
  a redundant persist is elided only when both the epoch and the head
  match, which also enables recording proven-no-op skip outcomes behind
  an elided persist at the same post-persist hook. `hard_reset` bumps
  the reset epoch FIRST and then purges the resident entity map under
  the map's own lock (one entity id lives in exactly one system: every
  resident at reset time belongs to the dead system; the purge also
  fixes the pre-existing dead-weak-entry accumulation), so a persist
  straddling a reset stamps a dead marker and stale residents are
  unreachable from ingest; `create()` and `join_system`'s raw set_states
  stay documented safe bypasses (marker and applied-set lag storage,
  never lead it). Reds: five GClock pins verified failing at the red
  commit (a covered-parent mis-stamp rejected against the
  materialization with zero payload reads; the sqlite and postgres
  round-trips of the annotation through set_state plus rehydration,
  multi-tip u32::MAX sentinel included; the motivating
  bodiless-adoption-then-commit case with zero peer event fetches; a
  durable node rejecting a lying wire annotation; and its structural
  companion, an annotation that does not cover exactly the head
  rejected as malformed on every node flavor); the pre-existing
  R-D2-2a stamp-equivalence oracle re-passed green through the rewired
  stamp (never red, the correct shape for a behavior-preserving
  rewire), sled's round-trip was green from birth as its commit
  discloses, and the indexeddb round-trip binary landed with the fix
  commit (verified in headless Chrome). Then R-D2-4a (zero
  set_state on a current no-op redelivery, counted at the storage
  boundary by an instrumented engine), the two R-D2-4b replacement pins
  per the 2026-07-09 reframe ruling (a persist straddling the epoch bump
  never yields a trusted marker, pinned at the mechanism seam and
  end-to-end across a real hard_reset with the persist held open at a
  storage gate; and post-reset the entity map is empty with old ids
  unreachable while held views stay frozen), and R-D2-4c (the ef68e081
  two-lane interleaving constructed fresh: lane B must persist while
  lane A's persist is held open, because the marker still names the
  pre-A head). Lane benches re-measured against the M2 table: the
  materialization erases most of the M2 stamping-plus-verification
  regression (single_writer_commit from +36.7% over pre-D2 down to
  about +5%, bridge_catchup back to pre-D2 levels, commit_to_subscriber
  from +38.5% to about +14%); the numbers live in the M4 milestone
  comment on #269. Post-milestone review remediation (ten items,
  commits ff9807d3..bb82d261, addendum comment on #269): same-entity
  persists are now serialized on a per-id lock, hard_reset is fenced
  against in-flight persists and materialization, durable annotation
  adoption and the indexeddb generation decode fail loudly on what they
  cannot inspect, and the R-D2-4c and straddler pins are amended to the
  fenced, serialized semantics.
- [x] **M5 comparison accelerations plus the immunity oracle.** The first
  milestone allowed inside the core/src/event_dag comparison machinery, on
  the sanctioned surfaces only; verdict-core semantics untouched (the
  census stands: no generation value reaches a verdict path, and the new
  counters are write-only during the walk by rule, so the stats surface
  cannot become a covert verdict channel). Landed red-first at every
  behavior change with the failing output in the red commit bodies:
  requested-id keying of the walk's bookkeeping including the accumulator
  LRU (a doctored payload served under the original id now completes with
  the true verdict instead of stalling to BudgetExceeded; content
  addressing remains the integrity mechanism at admission and at rest; a
  served-id mismatch is counted, never errored); CompareStats
  per-comparison counters (events fetched, precheck suppressions, edge
  check violations, demotions, id mismatches, budget decrements) on
  ComparisonResult, aggregates only, with the at-most-two-decrements
  budget invariant pinned by its own test; P1/P2 generation prechecks
  wired SUPPRESS-ONLY (operand GClocks read from the resident
  materialization under the head lock; eligibility computed at
  consumption: event in hand or a caller tip, not admitted-unverified,
  not the u32::MAX sentinel, not walk-demoted; a precheck can only
  suppress a positive fast-path attempt, and the general walk decides
  everything); frontier ordering as the existing level drain with
  in-level max-eligible-generation-first order, EventId tiebreak,
  unknowns last (the per-entry priority expansion was NOT taken; no
  concrete case arose, and level semantics keep the chain-completeness
  law schedule-independent); walk-time edge checks that warn plus demote
  and never reject (gen(child) equals 1 + max over parent generations,
  saturating arithmetic, checked opportunistically where the walk holds
  the family; demotion is per-comparison ineligibility, admission stays
  the rejection boundary); canonical chain emission via sorted-input
  topological sort with a permutation pin (the Kahn sorter's tie order
  derives from input order, so the input is what gets canonicalized).

  THE M5 DEFECT, found by the coordinator after the main M5 ladder
  landed and fixed inside the milestone: the walk-time edge checks as
  first landed re-registered a blocked child under every still-missing
  parent on every re-check, so duplicate registrations multiplied per
  arrival on a wide antichain (2 to the width power). One compare() call
  on a width-64 antichain of 66 events exceeded 11 GB RSS and cargo
  bench OOMed a 128 GB machine twice before reporting anything. Fixed by
  an at-most-once pending-registration guard: edge_waiting is now
  bounded by the DAG's edge count, pinned by a unit red driving the
  exact killing arrival order (250 registrations against the 128-edge
  bound by the third arrival, failing fast with no memory risk), with
  all edge-check semantics preserved (one evaluation per child ever,
  warn plus demote never reject, opportunistic misses on cache eviction
  stay acceptable). The corpus blind spot that let nineteen commits of
  soundness testing miss it is closed: the immunity oracle gains the
  GC-WIDE-ANTICHAIN named seed (width 64, three pinned pairs including
  a width-64 operand clock) and a wide randomized tier
  (full-current-tip joins over 24..=48 events, GEN_ORACLE_WIDE_SEEDS,
  with the band-wide max parent degree asserted at least 8 so the tier
  cannot silently degenerate back to narrow corpora), plus the nightly
  scale knob.

  ORACLE ARMS (the M5 capstone), all green at the tip: (i) accelerated
  verdicts equal the brute-force reachability oracle on every immunity
  baseline (compare_with) plus the pre-existing 300-seed property arm
  (compare); (ii) the two-channel gen-corruption immunity matrix (lying
  retriever, lying operand GClocks, cross-mismatch where the channels
  disagree): at gate defaults 300 randomized corpora of 5..=10 events,
  4 clock pairs each, 8 corruption shapes (gc-rand-payload,
  gc-rand-gclock, gc-rand-cross, gc-deflate-meet, gc-inflate-tip,
  gc-sat-interior, gc-sat-tips, gc-genesis), plus 20 larger corpora
  (20..=40 events) and 10 wide corpora (24..=48 events, full-tip
  joins), plus the named deterministic pins: GC-SEED-41 (the
  derivations 4.1 counterexample, baseline meet EXACTLY c1 plus m),
  GC-SEED-43 (the 4.3 self-refutation, meet EXACTLY f),
  GC-WIDE-ANTICHAIN (all three pairs), GC-SAT interior (contained AND
  seen: the violation counter is asserted) and tips (both channels at
  MAX disable the prechecks without a suppression), GC-GENESIS (roots
  claiming 0, 7, or MAX steer neither roots nor verdicts), the
  cross-mismatch pin (channels contradict, P1 fires, nothing BOUND
  moves), and the GC-CONTROL canary (structural corruption IS detected,
  so a green matrix is not vacuous). Every corrupted run is
  BOUND-identical to its honest baseline (verdict variant, chain
  semantic core with the completeness law, meet, immediate-children
  sets, layer partition); budget is 4x corpus with BudgetExceeded
  anywhere treated as a harness bug; failures carry one-line ORACLEFAIL
  artifacts with a format pin. (iii) synthetic saturation, distinct
  from GC-SAT: an honestly saturated corpus preserves verdicts through
  the disable path with zero suppressions and zero violations. (iv)
  every honest stamp equals its brute-force depth on every immunity
  baseline. The nightly immunity job scales to 25k randomized plus 2k
  large plus 1k wide seeds under dedicated knobs. The STOP-AND-REPORT
  discipline held: no BOUND divergence was observed in any run at any
  point in the milestone.

  Bench evidence (caveat: ambient load 5 to 10 from unrelated user
  applications throughout; every comparative claim below is a same-day
  paired A/B of identical bench sources at the M4 tip bb82d261 versus
  this tip, which cancels the environment): per-lane medians are within
  noise across the suite (commit_to_subscriber +3.0%, single_writer
  -0.7%, bridge_catchup +0.1% and -0.1% at 1000 and 5000, fresh_fetch
  within plus or minus 5%), except subscription_establishment/1000,
  whose run-to-run scatter today (plus or minus 14% at BOTH tips,
  overlapping bands) makes its single +21% pair reading inconclusive;
  the M2/M4 standing watch item on that lane's unexplained improvement
  stays OPEN pending a genuinely quiet machine (today the improvement
  does not reproduce, and the environment confounds the comparison).
  The event_dag micro suite at this tip is recorded as the E2
  reference (criterion means from estimates.json, table in the #269 M5
  comment): compare/wide_antichain/64 completes at 101 us where the
  pre-fix build detonated, compare cases sit 7 to 30 percent below the
  phase-2-start BASELINE.md medians (the prechecks and ordering shave
  the general walk), toposort pays 4 to 11 percent for canonical
  emission, and the applied-set cap-thrash cliff is characterized
  empirically (zero fetches inside the applied-set window, roughly
  twice the depth at the first evicted position, saturating at corpus
  size; the production cap of 65536 sits above the largest expected
  bridge batch; outcome never changes, only cost). The end-to-end
  eviction and rehydration pin holds: the GClock survives eviction plus
  rehydration byte-identical with ZERO get_events calls at the storage
  boundary, on single-tip and multi-tip unequal heads. All bench runs
  ride scripts/memwatch.sh, an 8 GB process-tree RSS watchdog added
  this milestone.
- [x] **M6 close-out.** The kill-switch: one node-level runtime flag
  (`Node::set_generation_accelerations_disabled`, homed on the
  acceleration-eligibility context `UnverifiedEvents`, the one node-level
  object every comparison site already receives) makes all three
  generation consumers dormant: the P1/P2 prechecks are skipped, the
  level drain runs in plain id order, and the walk-time edge checks are
  not evaluated at all, while STAMPING AND ADMISSION VERIFICATION STAY
  ON. The bridge walk (collect_event_bridge) moved from the bare compare
  wrapper to compare_with carrying the node's eligibility context, so
  the switch and D2-4 eligibility reach that walk too. Red-first: two
  dormancy reds verified failing at their commit (zero precheck
  suppressions and zero edge-check equation evaluations with the switch
  thrown, on a shape whose enabled twin provably shows both, through
  both the direct comparison seam and the apply_event operand path,
  observed via the new write-only CompareStats.edge_checks_evaluated
  counter); the equivalence pin (bound observables byte-identical to
  the enabled world across forty narrow plus five wide randomized
  corpora plus the named seeds, honest AND corrupted, with total
  dormancy) and the stay-on integration pin (switches thrown on both
  nodes: genesis stamps 1, its child stamps 2, a forged claim of 5
  rejects typed with expected 3, the honest twin applies) are green
  from birth with the justification and hostile-edit evidence in the
  fix commit body (swapping the walk's sides under the switch reddens
  exactly the equivalence pins; gating the executor's check_generation
  on the switch reddens exactly the stay-on pin). The indexeddb
  event-column generation read (read_generation in
  storage/indexeddb-wasm/src/collection.rs) is brought up to the M4
  remediation's checked loud-failure standard through a shared proto
  helper (proto::wasm::decode_generation, now the ONE conversion rule
  for both indexeddb generation read paths), red-first in the headless
  Chrome suite (NaN decoded to generation 0 at the red commit); the
  pin's dump arm also exposed and fixed a from-birth runtime defect in
  dump_entity_events (its key range used the wasm_bindgen class handle,
  never a valid IndexedDB key, so every call failed before reading).
  Threat-model addendum lands in specs/concurrency/threat-model.md
  (section 3.8: inflation bounded at honest-walk cost, demote-not-
  reject with the cannot-fire-under-honest-operation operational note
  and #335 as the escalation home, the precise ephemeral wedge posture,
  the durable never-adopt-uninspected posture; C4-22 rewritten from
  enforced-by-absence to as-built; gaps G-10 reset residue and G-11
  fetch_states row drop). The internals book and doc-comment residue
  from the M4 coherence panel are corrected (event-dag.md's Event
  definition and a new generation-accelerations section,
  storage-engines.md's engine matrix with the D2 columns,
  entity-lifecycle.md and compare-apply-cycle.md carrying the
  applied-set skip and marker elision, GLOSSARY.md pointing at the
  tip-of-tree internals, and the stale present-or-future-tense
  comments in proto and core flipped to as built). phase-2.md:
  workstream D item 2 rewritten to the as-built reality, checklist D2
  ticked, and the parking lot gains the maintainer-framed post-scope
  volume test suite item. The D3-gate agenda note is posted as #271
  comment 4948578287 (generation distance feeding trust or policy
  awaits the gated discussion; #335 parked alongside).

## Close-out evidence (M6)

**Oracle inventory, gate defaults (all green at the close-out tip).**
Arm i, accelerated verdicts equal brute-force reachability: asserted on
every immunity baseline (compare_with) plus the pre-existing property
arm at 300 seeds (compare). Arm ii, the two-channel gen-corruption
immunity matrix (core/src/event_dag/tests.rs::gen_corruption_immunity):
300 randomized corpora of 5 to 10 events, 4 clock pairs each, 8
corruption shapes per pair where meaningful (gc-rand-payload,
gc-rand-gclock, gc-rand-cross, gc-deflate-meet, gc-inflate-tip,
gc-sat-interior, gc-sat-tips, gc-genesis), plus 20 large corpora (20 to
40 events) and 10 wide corpora (24 to 48 events, full-current-tip
joins, band-wide max parent degree asserted at least 8); named
deterministic pins always on: GC-SEED-41, GC-SEED-43, GC-WIDE-ANTICHAIN
(three pairs at width 64), GC-SAT interior and tips, GC-GENESIS, the
cross-mismatch pin, and the GC-CONTROL canary that proves the harness
detects a genuine structural divergence. Arm iii, synthetic saturation.
Arm iv, every honest stamp equals its brute-force depth. Nightly scale
(.github/workflows/nightly-oracle.yml): 25000 randomized plus 2000
large plus 1000 wide seeds under the GEN_ORACLE_SEEDS,
GEN_ORACLE_LARGE_SEEDS, and GEN_ORACLE_WIDE_SEEDS knobs, alongside the
pre-existing 100k-seed property oracle. New at M6: the kill-switch pins
(two dormancy reds; the equivalence arm driving forty narrow plus five
wide corpora, two pairs each, honest and corrupted, plus four
named-seed runs through the same bound-observable protocol; and the
end-to-end stay-on pin in tests/tests/generation_admission.rs).

**Bench evidence.** Recorded in the #269 milestone comments and not
re-measured at M6 (M6 adds no hot-path work: the kill-switch is one
atomic load per comparison, default off). The paired A/B lane table
(M4 tip versus M5 tip, same day, same sources) shows the comparison
machinery costs nothing measurable on the production lanes, and the
event_dag micro suite at the M5 tip is the recorded E2 reference with
compare cases 7 to 30 percent below the phase-2-start BASELINE.md
medians: both tables with methodology and caveats in #269 comment
4942109007 (the M5 record). The M4 lane recovery table
(single_writer_commit from +36.7 percent over pre-D2 down to about +5,
bridge_catchup back to pre-D2, commit_to_subscriber from +38.5 to
about +14) is in #269 comment 4932828930. The
subscription_establishment quiet re-measure stays an open watch item
for the start of E2.

**Gate at the close-out tip (92317f65), all eight legs run and green.**
The three-package suite with --no-fail-fast (81 test binaries, zero
failures); cargo check -p ankurah-core --features wasm; cargo check
-p ankurah-storage-indexeddb-wasm --target wasm32-unknown-unknown;
cargo check --benches -p ankurah-core -p ankurah-tests --features
ankurah-core/bench-internals; cargo fmt --all --check; taplo fmt
--check; the criterion smoke under the memory watchdog
(scripts/memwatch.sh 8192, every bench case exactly once, 26 of 26
Success including the formerly detonating compare/wide_antichain/64);
and the full indexeddb browser suite in headless Chrome (15 test
binaries, zero failures; required this milestone because M6 touches
the indexeddb read path).

**The two defect-and-remediation cycles, honestly summarized.** First,
the M4 review batch (db396bb8..bb82d261, fifteen commits): the review
fleet found one major concurrency defect (concurrent same-entity
persists could invert at the engine, leaving storage behind the
resident while the marker elided redeliveries, violating the
marker-never-leads-storage invariant) plus a reset porosity family
(hard_reset could wipe while a persist was in flight, postgres would
recreate the dropped table and re-insert dead-system bytes, and the
entity load paths could resurrect dead residents), a durable-side
adoption hole (an unresolvable wire annotation tip fell through as
accepted), and an indexeddb decode that coerced instead of failing.
All were remediated red-first where constructible: per-id persist
serialization, a node-wide reset fence, the typed unresolvable-tip
rejection, the checked GClock decode, plus three mutation-proven
coverage pins and a docs batch. Second, the M5 edge-check cascade
(fixed inside the milestone, red e7588dce, fix ba1e5e52): the
walk-time edge checks as first landed re-registered blocked children
on every incomplete re-check, so registrations doubled per arrival on
a wide antichain; one compare() call on a 66-event width-64 DAG
exceeded 11 GB resident memory and cargo bench took down a 128 GB
machine twice. The fix bounds registrations by the DAG's edge count
(at-most-once pending guard); the blind spot that hid it is closed
structurally (the wide oracle tier and GC-WIDE-ANTICHAIN seed, the
criterion smoke joining the gate, and the memwatch wrapper on all
bench runs). Both cycles are narrated in full in the #269 comments
(4937098423 and 4942109007).

## Pre-merge items (deferred to the re-parent onto #318)

The protocol version handshake machinery lives in #307, which is not present
on this stack's base. These land immediately after the re-parent, tracked in
#266 comment 4909916652:

- [ ] PROTOCOL_VERSION bump, the `compatibility_is_exact_match` pin update,
  the version-history doc-comment, and red R-D2-1c (old-protocol peer
  refused at `register_peer`).
- [ ] Cargo minor version bump for the workspace crates.

## Gate

Full validation gate green at every commit boundary (red commits exempt
for exactly their own new tests): `cargo test -p ankurah-core -p ankurah-tests
-p ankurah-jwt-auth`, `cargo check -p ankurah-core --features wasm`,
`cargo fmt --check`, `taplo fmt --check`. The wasm-only indexeddb engine
additionally compiles under `wasm32-unknown-unknown`. Since M3 the gate
also requires benches to compile:
`cargo check --benches -p ankurah-core -p ankurah-tests --features ankurah-core/bench-internals`
(benches are outside the compile surface of `cargo test` and
`cargo check`, which is how the event_dag bench stayed broken from M1
until M3 repaired it; the feature flag is required because
`bench_support` is feature-gated and the bench target declares no
required-features, so the unflagged form cannot compile it). Whenever a
milestone touches the indexeddb engine (M2, M4, M6), the gate also runs
the full indexeddb browser suite headless in Chrome
(`wasm-pack test --headless --chrome storage/indexeddb-wasm`). Since M5
the gate also includes the criterion smoke
`cargo bench -p ankurah-core --bench event_dag --features bench-internals -- --test`,
which runs every bench case exactly once: the wide-antichain memory
defect detonated deterministically in that mode while the plain test
suite stayed green, so compiling benches is not enough and the gate now
executes them once. Bench evidence runs additionally ride
`scripts/memwatch.sh` (kill at 8 GB process-tree RSS and fail).


dnorman:
Collapsed into #318 (2026-07-11). The phase 2 stack is being finished and collapsed into one PR before its rebase onto the metadata work (#307), per the maintainer's sequencing direction. The branch phase-2-d1-ingest was fast-forwarded to this branch's tip 36157754, so #318 now carries the entire D1 plus D2 ladder as a single review surface, and, being based on main, puts the whole stack under CI for the first time. Nothing here was rebased, squashed, or force-pushed: the commits are byte-identical on #318, and this branch (phase-2-d2-generations) is left in place at the same tip.

This PR's description remains the detailed D2 ladder record: milestones M1 to M6, the per-milestone red-test inventories, the immunity oracle arms, and the close-out evidence. The final three commits of the ladder are the post-ladder codex follow-up remediation (fd5466d2 red pin, 93fee211 the canonical-instance persist refusal, 36157754 docs), recorded in the #269 addendum and summarized in #318's Workstream D2 section.

The pre-merge deferred items (PROTOCOL_VERSION bump plus the cargo version bump, ruling and tracking at #266 comment 4909916652) move to #318's checklist and land after its rebase.

PR #379, opened 2026-07-17
ankql: NOT keyword requires an identifier boundary

dnorman:
Any identifier starting with "not" (note, notes, notify) lexed as a unary NOT plus a mangled remainder, because NotFlag lacked the !IDENT_CONT boundary guard the other keywords carry. This guards it and pins identifiers-starting-with-not alongside the still-working bare keyword.

Carved out of #307 to shrink that review: this is an independent parser fix with no dependency on the property-metadata work. Once merged, #307 rebases over it and the commit drops out as already applied.

Verification: cherry-picks cleanly onto main; cargo check -p ankql is clean; cargo test -p ankql passes (60 tests, including the new boundary pins). CI completes the matrix.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved SQL parsing so the unary `NOT` operator is no longer confused with identifiers starting with “not” (e.g., `note`, `notes`, `notify`, `not-field`, `notа`).
  * Kept existing behavior working for standalone `NOT` expressions like `NOT (status = 'x')`.
* **Tests**
  * Added coverage to ensure “not*” prefixed identifiers parse correctly while valid unary `NOT` syntax continues to parse as expected.

### DIFF
--- a/ankql/src/ankql.pest
+++ b/ankql/src/ankql.pest
@@ -63,4 +63,8 @@ NotFlag = @{ ^"not" ~ !IDENT_CONT }
 EOF = { EOI | ";" }
 WS = _{ WHITESPACE }
 WHITESPACE = _{ " " | "\t" | "\n" | "\r\n" }
-IDENT_CONT  = _{ ASCII_ALPHANUMERIC | "_" }
+// The continuation set must equal Identifier's own: IdentifierNonDigit
+// admits hyphen and Cyrillic letters, so a guard narrower than the
+// identifier grammar would let a keyword prefix mangle identifiers like
+// not-field.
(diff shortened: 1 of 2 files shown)


dnorman:
Both CodeRabbit findings verified against the grammar and adopted in the follow-up commit: IDENT_CONT now equals the identifier continuation set (IdentifierNonDigit plus digits), and the regression test pins the full parsed operand with hyphenated and Cyrillic cases added. Note the deliberate scope widening: all six keyword guards (and, or, true, false, null, not) share IDENT_CONT, so the fix also corrects the same pre-existing gap for the other five keywords, always in the direction of an identifier winning over a keyword prefix.

PR #381, opened 2026-07-17
proto: peers negotiate an exact protocol version or part ways

dnorman:
Presence carries PROTOCOL_VERSION, and the handshake refuses an incompatible peer with a PresenceRejected message naming both versions, instead of letting mismatched codecs misread each other's shapes. register_peer becomes fallible so a rejected presence propagates as an error instead of a half-registered peer; every connector threads the refusal through; the websocket server bounds pre-establishment silence (INITIAL_PRESENCE_TIMEOUT); and a pre-versioning peer is detected by wire shape and refused the same way. The protocol_version integration suite pins accept, refuse, and timeout end to end.

On main this negotiation is trivially satisfied since all peers speak the same version. It is the machinery #307 leans on when it changes the serialized contract incompatibly (model-id envelopes, identity-keyed state operations, resolved ORDER BY terms, registration bodies; the peering module doc enumerates them): mismatched peers part ways at the handshake, and durable nodes upgrade first in a rolling deployment.

One consequence worth an explicit look before merging: peer compatibility on main changes immediately. The Presence message gains a field, so a peer built before this commit fails the handshake (detected and refused as pre-versioning) rather than interoperating. If main must keep talking to existing deployed peers until #307 lands, this PR should wait; if peers deploy together, it is safe now.

Carved out of #307 to shrink that review; once merged, #307 rebases over it and the commit drops out as already applied. Three small riders that live in later #307 commits came along to make it standalone: tokio-tungstenite and futures-util as test dependencies plus tokio's test-util feature (the new suite pauses time), and the sim harness's Presence construction gains the version field.

Verification: cherry-picks cleanly onto main; cargo check green across ankurah-proto, ankurah-core, ankurah-websocket-client, ankurah-websocket-server, ankurah-connector-local-process; protocol_version suite passes (5 tests) and the websocket suite passes (4 tests). The wasm websocket client is updated as well; the wasm CI leg covers it.

## Summary by CodeRabbit

* **New Features**
  * Presence handshakes now include an explicit `protocol_version`, and peers can receive/emit `PresenceRejected` on protocol mismatches.

* **Bug Fixes**
  * WebSocket handshake is stricter: initial presence is required before a timeout, legacy/unversioned handshakes are refused, and application traffic before negotiation is rejected.
  * Handshake-time peer registration failures now fail the connection cleanly (avoiding reconnect hot-loops); local-process peer registration rolls back on rejection.

* **Tests**
  * Added a protocol-version handshake test suite and expanded WebSocket/raw-socket coverage for timeouts, early-traffic rejection, and mismatch behavior.


dnorman:
All three CodeRabbit findings verified against the code and adopted in the follow-up commit, plus one test. The PeerMessage arm now refuses application traffic until presence negotiation completes, with a new pin (client_refuses_peer_message_before_presence) that drives a fake server leading with a PeerMessage; the pin was verified to fail without the guard. The wasm client now registers the peer before publishing Connected and rolls registration back if a concurrent transition claims the state first, closing the transient window where observers saw Connected with no registered peer. The test server helper now probes for positive listener readiness instead of the fixed 10ms sleep plus task-liveness check.

dnorman:
The follow-up finding is correct and adopted: set_state writes unconditionally and returns false only when the client is gone or no longer owns the connection, so the prior rollback left an orphaned connection reading Connected with no registered peer, and a later disconnect would deregister twice. The orphan now clears its state first (a plain write, since there is no owner left to notify), deregisters exactly once, and closes its socket. The misleading claimed-the-state comment is gone. Verified by reading handle_state_change: false comes from a dropped client or a stale connection identity, both meaning this object is unobserved.

dnorman:
Per maintainer direction, the protocol version is now 1, not 3: version numbers follow published releases rather than development steps. 0.9.x and earlier carry no version in Presence and classify as version 0 by wire shape; version 1 is the wire of 0.10.0, which is not yet released to crates.io. The constant doc in proto/src/peering.rs now states the numbering rule and records history by release. Landed as an additive commit on top of the restacked branch; happy to fold it into the base handshake commit instead if a rewrite is preferred.

PR #410, opened 2026-07-21
feat(jwt-auth): add issuer trust

dnorman:
## What

- Add native issuer-trust support to `ankurah-jwt-auth` behind the optional `issuer-http` feature.
- Discover and validate an OpenID Connect issuer, cache its RS256 JWKS, and verify strict access tokens synchronously.
- Replicate a versioned `IssuerTrustDescriptor` through `JwtPolicy` while preserving legacy PEM signing-key mode.
- Make descriptor-only ephemeral agents policy-ready without fetching or storing issuer JWKS.
- Document the public contract and add adversarial, rotation, policy-sync, and compatibility coverage.

## Why

The existing extension assumes the resource server owns a local JWT signing key. Resource servers using an external identity provider instead need one trust root: the configured issuer and its discovered JWKS. This moves authentication to strict issuer verification while keeping authorization policy-owned inside Ankurah.

## Security and behavior

- Require RS256, non-empty `kid`, `typ: at+jwt`, no `crit`, exact issuer, an allowed audience, `exp`, `sub`, and `token_use: access`.
- Discard token-supplied roles and install descriptor-owned default roles.
- Validate HTTPS discovery before I/O, reject redirects and non-success responses, require exact discovery issuer, and validate the discovered HTTPS JWKS URL.
- Use 5-second connect and 10-second request deadlines.
- Keep verification network-free with an atomically replaced last-good key cache.
- Bound unknown-`kid` refreshes with a shared single-flight gate and 30-second post-refresh cooldown.
- Reject empty/duplicate key IDs, unusable key sets, and RSA moduli below 2048 bits.

## Compatibility and impact

- Legacy local signing and verify-only PEM paths remain supported.
- Existing stored `jwtpolicy` entities remain readable because `trust_json` is optional.
- Adding `trust_json` to the public `JwtPolicy` struct requires downstream Rust struct literals to initialize it.
- Native issuer HTTP support adds optional `reqwest`/rustls dependencies only when enabled.
- Subscription token-expiry enforcement is intentionally deferred to a separate phase.

## Validation

- `cargo test -p ankurah-jwt-auth --all-features`
- `cargo check -p ankurah-jwt-auth --no-default-features`
- `cargo clippy -p ankurah-jwt-auth --all-features --no-deps`
- `cargo fmt --all -- --check`
- `git diff --check`
- Independent code/security review: LGTM, no landing blockers

WASM validation is currently blocked by a pre-existing `ankurah-core/src/task.rs` cfg leak that reproduces without `issuer-http`; this patch keeps HTTP discovery and refresh native-only.


dnorman:
Restack note: this branch was rebased onto current main (03d0c327, the session substrate). The rebase itself was conflict free and bit identical to git's mechanical merge, but the full test battery surfaced a defect in the PR as authored: test_ephemeral_readiness_accepts_watcher_descriptor_without_http requires the watcher feature (its publisher lives behind cfg(feature = "watcher")) yet carried no gate, so it fails under the default feature set CI uses. Verified by control run on the unmodified PR head: default features 1 passed 1 failed, with --features watcher 2 passed. Applied the one line fix here: the test is now gated behind the watcher feature. With features watcher,issuer-http the full jwt-auth suite is 100 passed, 0 failed.

PR #423, opened 2026-07-26
[ci-probe] write lock alone on 6d6084fd - do not review

dnorman:
Temporary CI bisect probe for the read_your_writes_at_origin failure at 2a5db900 (fails on GitHub runners only, 3 of 3, while the parent passes 4 of 4). This branch is the passing parent plus only the generate_commit_event write-lock change. Will be closed and deleted after the runs.


dnorman:
Probe complete: this configuration (parent + write lock alone) failed read_your_writes_at_origin 2 of 2 fleet runs (seeds 1 and 0), isolating the generate_commit_event write lock as the trigger. Fix landed on metadata-02-catalog as cef2ff75 (read lock restored, membership drain in its own short scope).

PR #424, opened 2026-07-26
[ci-probe] MembershipSet with read lock restored - do not review

dnorman:
Temporary CI bisect probe for the read_your_writes_at_origin failure at 2a5db900. This branch is the failing tip with generate_commit_event back on a read lock (membership drain in a separate preceding write scope). Will be closed and deleted after the runs.


dnorman:
Probe complete: this configuration (MembershipSet tree + read lock restored) passed read_your_writes_at_origin in 2 of 2 fleet runs; its one red leg was the pre-existing server_edits_subscription flake (PR #299 lineage), unrelated to the branch. This probe's shape landed on metadata-02-catalog as cef2ff75.

PR #425, opened 2026-07-26
The catalog eats its own dogfood: typed row models and a self-feeding map

dnorman:
The catalog now uses Ankurah's ordinary typed model machinery for its own durable data. Catalog rows are generated Models, durable registration writes them through a normal Transaction and Mutable values, and the node's catalog projection is maintained by typed LiveQueries and Views.

This follows #422 and targets `main`.

## Why

The earlier implementation still treated catalog maintenance as a privileged parallel data path: registration planned and encoded raw events, and the durable projection consumed raw entity state. That undercut the point of dogfooding the model stack and duplicated behavior that Context, Transaction, Model/Mutable/View, LiveQuery, storage admission, and Reactor notification already own.

The only special concept this needs is authority. A durable node must be able to maintain its own reserved system models without pretending that infrastructure work belongs to an application principal.

### Typed catalog models

`SysModelRow`, `SysPropertyRow`, and `SysModelPropertyRow` are ordinary `#[derive(Model)]` types. Their built-in identities come from `#[model(system = "...")]`, so they never consult the catalog they describe.

The derive now has two independent controls needed by these internal models:

- `base = "crate"` changes the generated Rust path.
- `no_ffi` suppresses generated WASM and UniFFI surfaces.

Model attributes are parsed strictly: unknown keys and invalid forms are compile errors.

### Local SystemRoot authority

`Context` now distinguishes `Session(ContextData)` from `SystemRoot`.

SystemRoot:

- is not a session and carries no application credentials;
- is constructible only inside core and only for durable nodes;
- may operate only on built-in system models;
- bypasses the application `PolicyAgent` for local system reads, queries, and commits; and
- otherwise uses the same transaction validation, event application, storage, and Reactor notification path as an application context.

Ordinary application contexts still cannot commit to `_ankurah_` collections, and remotely submitted protected-collection commits remain rejected.

### Registration uses Models and Mutables

Durable `RegisterSchema` materialization now opens a SystemRoot Context and uses the normal typed Transaction/Model/Mutable APIs to create or update catalog rows. `RegistrationPlan`, raw LWW event construction, and the parallel catalog write path are gone.

The remote request is still authenticated before it is accepted. Once accepted, materialization is local infrastructure work rather than work performed under the caller's policy context.

Ephemeral nodes deliberately continue to use the `RegisterSchema` RPC and fold its response for now.

### The durable catalog projection uses typed LiveQueries

A durable node opens three SystemRoot LiveQueries—one for each catalog collection—and applies their typed Views to `CatalogMap`. Initial results warm the projection; later ordinary commits update it through the same subscriptions.

Malformed rows are logged at error level, omitted from the projection, and do not stop later valid rows from being applied. Reset/readiness fences still prevent stale work from surviving a system-epoch change.

### Stable identity hints

`UniqueStructId` and `UniqueFieldId` are deterministic source-derived identifiers carried in descriptors and over `RegisterSchema`. They intentionally reach the server but are not used for matching yet.

A follow-up TODO records the observability needed before using them operationally: asynchronous successful-binding telemetry with declared and resolved identities, labels, match basis, declaring crate package/version, and application/source revision when the runtime can provide it.

### Purpose-first source organization

Every Rust source file touched by this PR now begins with module documentation that states the file's purpose, domain nouns, and boundary with neighboring modules. The repository's coding standards require that intention-setting step for touched source files.

## Deliberately out of scope

- Session aggregation/challenge-response work in #426.
- Ephemeral catalog LiveQuery reads; after #426 they can use Session-based subscriptions while keeping the RPC for writes, races, and migration fallbacks.
- Server-side matching on the source-derived unique IDs.
- The per-system descriptor/property lookup once-cell optimization.

## Verification

- `cargo test -p ankurah-core` — 213 passed, 1 ignored; doc tests completed.
- `cargo test -p ankurah-tests` — full package passed, including compile-fail, websocket, reset, protection, registration, and catalog dogfooding coverage.
- `cargo check -p ankurah-core --all-features` — passed.
- `cargo check -p ankurah-core --features wasm --target wasm32-unknown-unknown` — passed; CI now runs this target-specific check.
- `cargo fmt --all -- --check` and `git diff --check` — passed.

Derive coverage also pins explicit-default base-path behavior and nested user-module unique-ID hygiene.

The focused catalog coverage includes typed executor rows, feed-only updates, malformed-row isolation/continuation, application-policy isolation, protected local and remote writes, restart/reset fencing, concurrent registration, explicit ID verification, and idempotence.

A full `cargo test --workspace` remains blocked by an existing JWT extension derive failure under workspace feature unification; clippy likewise stops on an existing denied lint in `core-types`. Neither failure is introduced by this PR.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -64,7 +64,7 @@ jobs:
         run: rustup target add wasm32-unknown-unknown
 
       - name: Check standalone core WASM feature
-        run: cargo check -p ankurah-core --features wasm
+        run: cargo check -p ankurah-core --features wasm --target wasm32-unknown-unknown
 
       - name: Run WASM tests
         run: ./scripts/run-wasm-tests.sh
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -63,6 +63,9 @@ jobs:
       - name: Add wasm32 target
         run: rustup target add wasm32-unknown-unknown
 
+      - name: Check standalone core WASM feature
+        run: cargo check -p ankurah-core --features wasm
+
       - name: Run WASM tests
         run: ./scripts/run-wasm-tests.sh
         env:
--- a/CODING_STANDARDS.md
+++ b/CODING_STANDARDS.md
@@ -1,0 +1,20 @@
+# Coding Standards
+
+## Source files state their purpose and vocabulary
+
+Every source file touched by a change must begin with a module-level
+documentation comment (`//!` in Rust, or the language's equivalent) that
+states:
+
+- the file's single overarching purpose;
+- the important nouns it introduces or owns and what each noun means; and
+- any easily confused responsibility that explicitly belongs elsewhere.
+
+Write or revise this block before changing the implementation. It is an
+intention-setting and alignment step, not a description invented after the
+code is finished.
+
+If the purpose and nouns cannot be stated clearly and concisely, stop and
+reorganize the code until the file has a coherent responsibility. Treat a
+missing, vague, or stale purpose block as a structural problem, and keep the
+block current when the file's responsibility changes.
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -173,6 +173,7 @@ dependencies = [
  "futures-signals",
  "futures-timer",
  "futures-util",
+ "getrandom 0.3.4",
  "indexmap 2.12.1",
  "itertools 0.14.0",
  "js-sys",
--- a/core-types/src/lib.rs
+++ b/core-types/src/lib.rs
@@ -10,6 +10,7 @@ mod entity_id;
 mod error;
 mod model_id;
 mod property_id;
+mod unique_id;
 mod value;
 mod value_type;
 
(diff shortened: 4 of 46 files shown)


dnorman:
Review directive from Daniel (2026-07-26, recorded during the PR #422 review): core/src/schema/registration.rs seems fairly horrific and should be looked at very closely. The whole catalog system should be examined closely for removal of code smell, and simplified and golfed quite a lot.

Context for whoever executes this: the deep golf of the catalog was deliberately held shallow until this PR's dogfood direction was proven, since the subscription-fed map replaces part of the warm machinery and polishing doomed structure would have been wasted work. That verdict is now in, so the golf belongs in this PR's orbit or immediately after it. Known starting points from earlier review notes: the registration executor's two-pass upsert and plan plumbing in registration.rs is the worst offender by his read; EnsuredSchemaBinding can likely shrink to (model, confirmed) with the field map recomputed from the catalog map; the warm and fence generation scaffolding in catalog.rs should be re-derived from what the subscription feed actually needs rather than kept by inertia.

PR #426, opened 2026-07-27
Live credential sessions: updatable ContextData, the credential union, and query status

dnorman:
**Stacked on #434.** The vocabulary and the live-source structure landed in the base PR #434 (sessions threaded everywhere a credential was baked: livequery construction, relay sends, gap fetchers, the union carry, the jwt equality contract), independently mergeable because standing queries there behave exactly as on main. This PR is the reactions layer: re-permission, denial as state, status, the effects queue, attempt identity, and the reactor liveness gates. Its tip tree is byte-identical to the pre-split branch, so the review record below covers the composed result.

A Context's credential becomes live, a query's credentials become plural, and a query's condition becomes observable. One `Session` per Context wraps the current ContextData in a signal; holders read the value at use time instead of freezing clones at construction; `Context::update_cdata` replaces the credential in place, so a token refresh or a re-login no longer means rebuilding the Context and every livequery under it. A query is credentialed by a `Sessions` source: one session (an ordinary context's query) or the node's whole live `SessionSet` (a system query reading under the union of active credentials, the shape the catalog manager's standing queries will use). And every livequery exposes a two-leg reactive `status()` that fails, un-fails, and re-fails as credentials, peers, and the server move, with credential denial reported as a status on set-backed queries and the predicate never altered. Targets main, independent of #425 (which is rebuilt over this after it lands). Two commits: the mechanism, then the proofs.

## What this adds

**`Session<CD>`, `SessionSet<CD>`, and `Sessions<CD>`** (core/src/session.rs). A Session is one live credential: interior-mutable current value plus change reactivity, both via the existing `ankurah_signals::Mut`. Sessions are strictly per-context and never deduplicated: two credentials that compare equal today are independent sessions that can diverge tomorrow, so deduplication belongs at the point of consumption, over current values. The node's SessionSet enumerates live sessions; membership is the `Arc<Session>` strong count itself, culled by RAII, and a live query extends its session past the Context handle's drop. The set is itself a first-class signal implementing the standard vocabulary (`Signal`/`Get`/`Peek`/`Subscribe`) at `Vec<CData>`, firing on registrations, culls, and member updates alike. `Sessions` is the credential source everything holds, one enum for the three places credentials come from: `One(Arc<Session>)` (an ordinary context), `Set(SessionSet)` (the system context, the node's whole live set), or `Held(SessionHolder)` (the server's per-query holder: a private set plus the memberships that constitute it, replaced wholesale on every re-validated subscribe, revoking before registering so an overlapping snapshot can see old, partial-old, empty, or new members but never a union). One `snapshot() -> Vec<CData>` and composite generations close the snapshot-then-listen window for every shape; only `One` can write. `Session::detached` covers infrastructure credentials that must stay out of the node's enumeration. The `ContextData` trait moves here, its natural home, behind a node.rs re-export.

**Read-through instead of frozen clones.** `NodeAndContext` snapshots its credential source once per logical operation (one credential set per commit, per fetch, per get, never mixed mid-operation); the gap fetcher and the relay's per-query `Content` hold the source and read the then-current values at each use, so a reconnect re-registration after a refresh carries the fresh credentials.

**The union rides the wire that was already shaped for it.** The request envelope has always carried `Vec<AuthData>`, and the PolicyAgent's read-side checks (`can_access_collection`, `filter_predicate`, `check_read`) were already generic over N credentials, with commits pinned to exactly one. This PR supplies what was missing: the live client-side source of N (the SessionSet), the server's SubscribeQuery arm (it demanded exactly one credential; it now validates all and evaluates the union), a per-query server-side holder (the `Sessions::Held` arm, session.rs) whose replacement discipline fails closed in the revocation direction, and the one real defect the N-shaped surface hid: jwt's `filter_predicate` took the FIRST authenticated context's claims when scope rules existed. It now ORs the per-context narrowings of the contexts authorized for the collection (an authenticated context without the collection's read privilege contributes no branch, and a set with none is denied), mirroring `enforce_read_scope`'s existing per-credential evaluation. A vec of one reproduces today's behavior exactly, everywhere.

**The system context.** `Node::system_context()` (pub(crate)) builds a Context whose credential source is the node's whole SessionSet: its queries read under the union and re-permission as sessions come, go, and refresh. Write operations, registration, and `update_cdata` refuse through it, because those act as one principal.

**`QueryStatus`: two legs, nothing latches.** Every livequery exposes `status()`, one reactive value. The local leg says what this node's reactor serves: `Pending` (no registration has landed yet), `Denied { reason }`, or `Active { version }`. `Active` is only ever reported by a registration that actually landed (the reactor's pre-notify hook is its sole writer, version-floored), so the local leg cannot claim service that never installed; an activation failure leaves the last truthful state standing with the error latched on `error()`. The remote leg says where the upstream subscription stands: `None` on durable nodes, `Pending`, `Requested`/`Established` per version, `Denied` for a local refusal to sign, `Error` carrying the server's refusals and other non-retryable failures. The relay reports every transition through the existing subscriber callback, each report stamped with a sequence drawn under the relay's lock and floored at the livequery, so out-of-order delivery can never leave the public status stale. No signal fires while a livequery lock is held: status and selection dispatch through an ordered effects queue drained outside the locks (single drainer; a re-entrant enqueue is picked up by the active drainer), and resultset broadcasts ride their write guard past release. A status or resultset subscriber may therefore call `update_selection` from inside its callback instead of deadlocking, which the reentrancy test pins. Completions and failures are accepted by ATTEMPT IDENTITY: every send stamps `Requested(peer, version)` and bumps a monotonic attempt id under the relay's lock, and a report acts only when it matches all three (`(peer, version)` alone is reusable, because a reconnect of the same peer re-sends the same version), so a dead attempt's late echo (superseded, re-routed after its peer disconnected, already resolved, or a prior try of the same version) can neither claim Established, re-point a live subscription, nor knock one back to pending; an entry sitting at pending gets a kick so a lost dispatch cannot strand it. Retryable transport failures report as `Pending` while the retry task re-attempts; a credential change re-kicks a failed leg through the normal versioned update flow, so recovery is observable rather than silent. The previous reporting structurally could not express recovery: the relay's per-query state was invisible outside a test hook, and the livequery's error signal was write-only, seven writers and no clearer (it now clears on successful activation, so `error()` and `status()` agree after a failure-then-recovery).

**Denial is a state, not a failure.** With zero eligible credentials a set-backed query constructs successfully, holds no reactor registration, serves nothing, reports `Denied`, and retains the user's intent untouched. And for EVERY query kind, a credential change that revokes a standing query's access (a re-login to a weaker subject, a logout implemented as update-to-anonymous) flips it to `Denied` and claws the rows back out of the live resultset (locally persisted rows are untouched), rather than freezing it serving the old credential's data. The intent still goes upstream for the server's own verdict, so asymmetric policies where the server grants what the client denies keep working. Revocation is enforced on BOTH sides: when a re-validation fails, the server removes the standing registration too (floored on the request's version, so a delayed older request that now fails validation cannot tear down the live newer registration), and nothing keeps streaming to a denied subscription under the credential it replaced. A tombstone against an older request re-registering after removal is follow-up work, tracked. Revocation also wins against in-flight work: the authoritative selection and denial state live under one lock that activation brackets (snapshot before, re-validate after the reactor call), so an activation that raced the denial re-asserts the claw-back instead of leaving a zombie registration; asynchronous gap fills re-check under the reactor's state lock that their query still exists at the version they fetched for before applying, so a fill that started under the old credential dies instead of repopulating a clawed-back resultset. The next credential change re-derives: local activation re-enters through the first-activation path at the CURRENT version (so waiters complete), the relay re-attempts, and both legs heal. Two loudness rules survive by design: single-credential queries fail construction loud when they never had access, and an explicit `update_selection` under a denying credential errors loud, because in both cases the caller is present to hear it.

**Re-permission re-derives, never re-sends.** The livequery retains the user's PRE-FILTER intent, and one refilter closure is the single derivation authority: policy checks and filtering run under the credentials current at derivation time, for creation, selection updates, and every re-permission alike. A credential change can therefore never re-send a filter baked under old credentials (which went permanently dark under client scope rules, the deployed policy-sync shape). The change listener lives on the livequery itself, so a DURABLE node's local queries re-filter their reactor registration, and relayed queries additionally re-send upstream through the existing versioned update flow, zero new protocol. The server's subscribe upsert re-validates whatever arrives and re-narrows the selection under the new claims.

**Version integrity.** Read-refilter-mint-store is one atomic step under the livequery's update lock, and the server's `update_query`, the relay's stored content, and the relay's status transitions all enforce version floors, so racing updates cannot regress a standing query to an older credential's selection.

**Change-gated delivery on operational identity.** `ContextData`'s `Eq` means operational identity: values compare equal only when substituting one for the other changes nothing observable, token included. (The jwt impl previously compared only the subject, an accident of serde_json hashability; equality and hash are now full-value, with the custom-claims map hashed in sorted key order.) `Session::update` gates on it, so subscribers hear exactly the updates that changed something: a token refresh carries a new token and re-permissions; an identical update is a complete no-op. Unit tests pin both directions.

**Deleted: `predicate_context`.** A node-level query-to-credential map written on subscribe, removed on unsubscribe, and read by nothing.

## Verification

Full battery 507 passed / 0 failed across 95 binaries on the main base, plus core's own unit suite (221). What the suites prove:

- **tests/tests/session_cdata.rs** drives the context surface over a real two-node pair: an equal-comparing `update_cdata` is a complete no-op that leaves the standing subscription untouched (the value-different path runs end to end under real credentials in the jwt suites); sessions extend through livequeries and cull by RAII; a mismatched ContextData type is refused.
- **extensions/jwt-auth/tests/session_refresh_tests.rs** proves visibility follows the credential, with exact-id assertions: re-login on a relayed subscription re-scopes it server-side; the same holds with scope rules mirrored onto the client; a durable node's local livequery re-filters with no relay involved; one update re-permissions every standing query under the context, each through its own version authority. `logout_claws_back_the_resultset` pins revocation end to end: logout empties the live resultset and reports `Denied`; a commit the old credential could see does not stream into the denied subscription (the server tore down its side too); re-login heals with bounded initialization and recovers both rows, the one clawed back and the one committed while denied.
- **extensions/jwt-auth/tests/system_query_tests.rs** drives the full set-backed arc against a scoped server, with exact-id assertions: denied at zero sessions (local `Denied`, the server's refusal on the remote leg), healed to alice's slice by her login (a bounded `wait_initialized` pins that healing initializes at the current version), widened to the union of alice's and bob's slices by his, narrowed back by her logout, and returned to `Denied` by his. That arc is the catalog manager's standing-query lifecycle.
- **tests/tests/query_status.rs** pins the status machinery: the remote leg establishes per version and follows selection updates; durable-local queries report no remote leg; both legs report `Pending` on a disconnected ephemeral node; and re-entrancy holds from both directions (a status subscriber AND a resultset subscriber calling `update_selection` from inside their callbacks; each deadlocked before its respective dispatch rework).
- **Unit suites**: the jwt filter union (including the exclusion of authenticated-but-unauthorized contexts); full-value credential equality with an insertion-order-independent hash (context_tests); session.rs liveness culling, change-gated notification (a refresh fires, an identical update is a no-op), composite generations, and immediate revocation (a replaced member never appears beside its successor, even while a reader's Arc keeps the replaced Session alive).

## Not in this PR

A wasm-reachable `update_cdata` surface (the generic method cannot cross wasm_bindgen; browser token refresh needs a small typed wrapper in the bindings layer). A limit-gap-across-refresh test and a reconnect-after-refresh test (named coverage gaps, recorded). The catalog PR (#425, rebuilt over this) introduces the CatalogManager as a pure consumer of the system context; the LiveQuery/Context/peer_subscription layers are final for it. Durable-side expiry enforcement against standing subscriptions is deliberately out; this PR builds its prerequisites (per-query server-side credential holders and observable status).

## Review record

Fourteen independent review rounds ran against this branch during its build. Every confirmed finding is fixed in place; the load-bearing ones are pinned by the tests above.

1. Two agent review rounds (code quality, then code placement) during the initial build, plus CodeRabbit throughout; all nine CodeRabbit threads carry disposition replies.
2. A first Codex review produced the per-credential authorization gate on the filter union, the monotonic initialization floor, the version floor on server credential refreshes, and per-entity watcher cleanup on query removal.
3. An Opus review of the accumulated delta caught a self-deadlock the dispatch rework had introduced on the denial path (reactor deregistration under a held resultset guard; two tests hung until fixed) and a stale-activation hazard in a defensive re-registration. Both fixed: deregistration now runs before any guard exists, and a registration conflict falls back to the version-floored update path instead of ever removing a possibly-newer live registration. The same round hardened stale relay reports against clobbering an established subscription, re-armed gap filling when a fill is dropped as superseded, gated the error-latch clear on version currency, and made credential replacement's revocation explicit and immediate.
4. A second Codex review drove the final round: the server-side half of the claw-back; reactor broadcasts moved outside the subscription state lock; the initialization wait registers interest before checking its condition (a wake in the gap stored no permit and could strand a waiter); an unwind-safe effects drain; establishment guards on the third relay completion site, extended to same-version duplicates from since-disconnected peers; `Active` dispatching before the reactor's changeset notification; and the error latch versioned through the same ordered dispatch as the status.
5. An Opus review of that remediation delta then caught its over-corrections: the new relay guards made Established terminal while a dead peer's late completion could still create it (a silent permanent stall), and the server claw-back lacked a version floor (a delayed older denial could tear down its live successor). The repairs replaced the guard stack with attempt-keyed acceptance (the structural half of the request-identity suggestion, built on the Requested state every send already stamps), floored the server removal, swept three more broadcast-under-lock sites (a reactor-notification subscriber revoking the credential from inside its callback deadlocked; now pinned), moved error-latch enqueueing under the currency lock, and replaced two tests that passed without the fixes they claimed to pin: the logout proof now asserts the denied-window commit is absent from client LOCAL storage, the only observation that distinguishes the server's teardown from the client's own claw-back. The same round folded the server-side holder into the `Sessions` enum as its `Held` arm (review feedback on an aliasing field pair). The tombstone half of request identity remains follow-up work, #429.
6. A paired round then ran an Opus review and a Codex review over that repair delta at once, and they converged independently on the same two blockers. First, attempt identity was still incomplete: `(peer, version)` is reusable, because a reconnect of the same peer re-sends the same version, so a dead attempt's echo could still resolve its successor's state; every send now also bumps a monotonic attempt id under the relay's lock, and completions and failures must match it exactly. Second, the deferred membership writes that `evaluate_changes` broadcasts after releasing the state lock carried no liveness check, so a subscriber revoking the credential from inside an earlier write's broadcast left a later same-batch write repopulating the denied query's clawed-back resultset; each write now carries its query id and decision-time version and re-checks, under a re-taken state lock, that the query still exists at that version before applying, the same rule the gap fills already follow. The pin for that gate was verified to fail with the gate removed. The same round released the global subscriptions mutex before per-subscription system resets (a subscriber revoking synchronously re-enters query removal, which needs that mutex), widened the server claw-back floor to cover the window between a newer subscribe's credential seed and its selection update, floored the relay's error latch on the same report sequence as its paired status, and fixed the logout pin's wrong collection id, whose absence assertion had been vacuous against an on-demand-created empty collection; a positive post-heal arm now self-validates the name. The watcher-lifecycle race both reviewers also flagged predates this branch and is tracked as #430.
7. The round-10 delta got the same paired treatment, and CI itself contributed a finding: the corrected logout assertion, no longer vacuous, exposed its own test's synchronization, which waited on the client's instant local denial while the server's re-validation was still in flight, so on a slow runner the commit streamed through the not-yet-removed registration; the test now waits for the server's refusal to land on the remote leg, which the server sends only after its claw-back completes. The Opus review found that the round-10 deferred-write gate collapsed removal and supersede into one discard verdict, and a superseded ADD has no other recovery path (the superseding update only re-adds entities from a fetch that predates the batch), so a row belonging to the current selection went silently missing, precisely in the token-refresh case where the version bumps and the selection does not; the gate is now three-way like the gap fills', re-deciding superseded writes against the current selection under the re-taken lock, with a reactor unit test pinning the interleave (verified to fail with the re-decision removed). The Codex review found the attempt nonce guarded the relay's state transition but not its side effects (both completion arms activated the livequery before the attempt check, so a dead attempt's late completion could activate the local leg and clear an error a live attempt had latched; both arms now claim the completion first, and only the claiming attempt activates), that (query id, version) is not registration-incarnation identity (all three liveness gates now also require the registration to hold the same resultset instance the decision captured), and that a revoked registration's batch still reached changeset subscribers as membership adds (the livequery now drops deliveries after its own denial, pinned by a changeset observer added to the batch test and verified discriminating). Dropped writes cancel their watcher changes with them. The in-flight-request teardown facet and the local-denial persistence window are recorded on #429.
8. Rounds 12 and 13 attempted to also fix the CONCURRENT-ACTIVATION substrate the reviews had begun mapping (a narrow denial-heal window where a stale activation can resurrect old-credential content until its successor lands, and the decision-versus-application divergences around deferred writes, watcher changes, and item claims). Both attempts were themselves found defective by the next paired review: the round-12 claw-back arm was proved, with a deterministic fetch-delaying probe, to take down the successor registration it was meant to spare, and the round-13 repair closed that while opening watcher and lifecycle corners of its own. Rather than a fourth patch round, both commits were REVERTED and the branch retreated to its round-11 waterline, where the paired verdict had been clean: the substrate work is scoped out of this PR deliberately, and everything the four review rounds established about it (the enumerated windows, the failed approaches and exactly how each failed, the lock-order constraint, and the application-derived-truth design both review tracks converged on) is consolidated in #431 for a dedicated follow-up. Two durable artifacts of those rounds stay in the tree: the deterministic delay-engine test seam and the racing-activations convergence pin (tests/tests/activation_races.rs), which passes at the waterline and guards against any future superseded-path cleanup taking down its successor.
9. The maintainer's guide-driven review (2026-08-02) produced a language ruling and a semantics redesign. The ruling: doc comments must lead with what a thing exists for, and the Session generation docs narrated their reader's protocol instead; they were rewritten purpose-first. The redesign: the maintainer challenged notify-on-every-update, and archaeology showed JwtContext's subject-only equality was an accident of hashability (serde_json::Value implements no Hash, and the workaround narrowed Eq along with Hash when only Hash needed it). Equality and hash are now full-value, ContextData documents Eq as operational identity, and Session::update gates delivery on it: a token refresh compares unequal and re-permissions, while an identical update is a complete no-op. The equality pins flipped to match (context_tests, the session unit suite, the session_cdata no-op pin); the generation counter survives on a new justification (a change-and-revert between a livequery's snapshot and its listener installation leaves the value equal while an in-window send may have shipped the interloper).
10. The maintainer's per-layer minimalism audit (2026-08-03), after the branch split into a stack: each layer justifies its own contents. The generation counter, which had landed in the base PR with no consumer there, was deleted from both layers; the livequery constructor now closes its derive-then-subscribe window by comparing session snapshots under operational-identity equality, with the accepted change-and-revert corner documented at the site. The set-backed system context constructors, consumer-less even in the composed tree, were deferred to the catalog PR that first consumes them. Sessions::subscribe_changes moved into this PR beside its listener, and policy call sites inline their snapshots.

## Summary by CodeRabbit

* **New Features**
  * Live queries now support dynamic credential updates without recreating subscriptions.
  * Added live query status reporting for local and remote connection states.
  * Multi-session access now combines authorized visibility across active credentials.
  * Queries automatically recover when access is restored and show denied states when authorization is revoked.

* **Bug Fixes**
  * Improved protection against stale updates, activation races, duplicate notifications, and re-entrant subscription issues.
  * Prevented unauthorized changes from reaching client result sets.
  * Improved remote subscription retry and disconnect handling.

### DIFF
--- a/core/src/context.rs
+++ b/core/src/context.rs
@@ -298,18 +298,6 @@ impl Context {
         Self(Arc::new(NodeAndContext { node, sessions: session.into() }))
     }
 
-    /// A context reading under the union of the node's live credentials,
-    /// reactively (the credential source is the node's whole SessionSet).
-    /// For system queries such as the catalog manager's; not a user
-    /// surface. Write operations, registration, and update_cdata refuse
-    /// through it, because they act as one principal.
-    pub(crate) fn new_system<SE: StorageEngine + Send + Sync + 'static, PA: PolicyAgent + Send + Sync + 'static>(
-        node: Node<SE, PA>,
-    ) -> Self {
-        let sessions = crate::session::Sessions::Set(node.sessions.clone());
-        Self(Arc::new(NodeAndContext { node, sessions }))
-    }
-
     /// Replace this context's credential in place: a token refresh or a
     /// re-login. Livequeries and pending operations keep running; each
     /// holder reads the new value on its next operation, and the relay
@@ -407,9 +395,9 @@ where
     SE: StorageEngine + Send + Sync + 'static,
     PA: PolicyAgent + Send + Sync + 'static,
 {
-    /// The single credential write paths require. The system context
-    /// refuses: commits, registration, and event attestation act as one
-    /// principal, and the union has no single principal to act as.
+    /// The single credential write paths require. Set- and holder-backed
+    /// contexts refuse: commits, registration, and event attestation act
+    /// as one principal, and a union has no single principal to act as.
     fn write_credential(&self) -> Result<PA::ContextData, AccessDenied> {
         match &self.sessions {
             crate::session::Sessions::One(session) => Ok(session.snapshot()),
@@ -444,7 +432,7 @@ where
             debug!("Node({}).get_entity found local entity - returning", self.node.id);
             let state = local.to_state()?;
             let entity_id = local.id();
-            self.node.policy_agent.check_read(&cdata, &entity_id, collection_id, &state)?;
+            self.node.policy_agent.check_read(&self.sessions.snapshot(), &entity_id, collection_id, &state)?;
             return Ok(local);
         }
         debug!("{}.get_entity fetching from storage", self.node);
@@ -452,7 +440,12 @@ where
         let collection = self.node.collections.get(collection_id).await?;
         match collection.get_state(id).await {
             Ok(entity_state) => {
-                self.node.policy_agent.check_read(&cdata, &entity_state.payload.entity_id, collection_id, &entity_state.payload.state)?;
+                self.node.policy_agent.check_read(
+                    &self.sessions.snapshot(),
+                    &entity_state.payload.entity_id,
+                    collection_id,
+                    &entity_state.payload.state,
+                )?;
                 let state_getter = crate::retrieval::LocalStateGetter::new(collection.clone());
                 let event_getter = crate::retrieval::CachedEventGetter::new(collection_id.clone(), collection, &self.node, &cdata);
                 let (_changed, entity) = self
@@ -467,11 +460,13 @@ where
     }
     /// Fetch a list of entities based on a selection
     pub async fn fetch_entities(&self, collection_id: &CollectionId, mut args: MatchArgs) -> Result<Vec<Entity>, RetrievalError> {
-        let cdata = self.sessions.snapshot();
-        self.node.policy_agent.can_access_collection(&cdata, collection_id)?;
+        self.node.policy_agent.can_access_collection(&self.sessions.snapshot(), collection_id)?;
         // Fetch raw states from storage
 
-        args.selection.predicate = self.node.policy_agent.filter_predicate(&cdata, collection_id, args.selection.predicate)?;
+        args.selection.predicate =
+            self.node.policy_agent.filter_predicate(&self.sessions.snapshot(), collection_id, args.selection.predicate)?;
+
+        let cdata = self.sessions.snapshot();
 
         // Resolve types in the AST (converts literals for JSON path comparisons)
         args.selection = self.node.type_resolver.resolve_selection_types(args.selection);
--- a/core/src/context.rs
+++ b/core/src/context.rs
@@ -33,7 +33,24 @@ where
     PA: PolicyAgent + Send + Sync + 'static,
 {
     pub node: Node<SE, PA>,
-    pub cdata: PA::ContextData,
+    /// The live credential source. An ordinary context holds exactly one
+    /// session (registered in the node's SessionSet); the system context
+    /// holds the whole set and reads under the union. Every operation
+    /// snapshots it once at entry, so an update lands between operations,
+    /// never inside one.
+    pub sessions: crate::session::Sessions<PA::ContextData>,
+}
+
+/// [`Context::update_cdata`] failure.
+#[derive(Debug, thiserror::Error)]
+pub enum UpdateCdataError {
+    /// The supplied value is not this node's ContextData type.
+    #[error("ContextData type mismatch: this node's policy agent uses a different ContextData type")]
+    WrongContextDataType,
+    /// The system context has no single credential to replace; its
+    /// member sessions update through their own contexts.
+    #[error("the system context's credentials update through their member contexts, not update_cdata")]
+    SystemContext,
 }
 
 #[async_trait]
@@ -66,6 +83,12 @@ pub trait TContext {
     async fn commit_local_trx(&self, trx: &Transaction) -> Result<Vec<Event>, MutationError>;
     fn query(&self, collection_id: proto::CollectionId, args: MatchArgs) -> Result<EntityLiveQuery, RetrievalError>;
     async fn collection(&self, id: &proto::CollectionId) -> Result<StorageCollectionWrapper, RetrievalError>;
+    /// Replace this context's credential (type-erased across the trait
+    /// object; [`Context::update_cdata`] is the typed surface). Holders of
+    /// the session observe the new value on their next operation. STANDING
+    /// queries keep their creation-time derivation; the liveness PR wires
+    /// re-permission.
+    fn update_cdata(&self, new: Box<dyn std::any::Any + Send>) -> Result<(), UpdateCdataError>;
 }
 
 #[async_trait]
@@ -74,7 +97,11 @@ impl<SE: StorageEngine + Send + Sync + 'static, PA: PolicyAgent + Send + Sync + 
         &self,
         schema: &'static crate::schema::ModelStructDescriptor,
     ) -> Result<proto::ModelId, crate::schema::registration::RegistrationError> {
-        self.node.catalog.ensure_schema_for_use(&self.cdata, schema).await
+        // Registration acts as one principal, like every write path.
+        let cdata = self
+            .write_credential()
+            .map_err(|denied| crate::schema::registration::RegistrationError::Retrieval(RetrievalError::AccessDenied(denied)))?;
+        self.node.catalog.ensure_schema_for_use(&cdata, schema).await
     }
 
     fn node_id(&self) -> proto::EntityId { self.node.id }
@@ -82,7 +109,9 @@ impl<SE: StorageEngine + Send + Sync + 'static, PA: PolicyAgent + Send + Sync + 
         let primary_entity = self.node.entities.create(collection);
         primary_entity.snapshot(trx_alive)
     }
-    fn check_write(&self, entity: &Entity) -> Result<(), AccessDenied> { self.node.policy_agent.check_write(&self.cdata, entity, None) }
+    fn check_write(&self, entity: &Entity) -> Result<(), AccessDenied> {
+        self.node.policy_agent.check_write(&self.write_credential()?, entity, None)
+    }
     async fn get_entity(&self, id: proto::EntityId, collection: &proto::CollectionId, cached: bool) -> Result<Entity, RetrievalError> {
         self.get_entity(collection, id, cached).await
     }
@@ -98,6 +127,10 @@ impl<SE: StorageEngine + Send + Sync + 'static, PA: PolicyAgent + Send + Sync + 
         if trx.alive.compare_exchange(true, false, Ordering::AcqRel, Ordering::Acquire).is_err() {
             return Err(MutationError::General("Transaction already committed or rolled back".into()));
         }
+
+        // One credential snapshot for the whole commit: a session update
+        // mid-commit must not mix credentials across its phases.
+        let cdata = self.write_credential()?;
 
         // Generate events from the transaction entities
         let trx_id = trx.id.clone();
@@ -152,7 +185,7 @@ impl<SE: StorageEngine + Send + Sync + 'static, PA: PolicyAgent + Send + Sync + 
             event_getter.stage_event(event.clone());
             forked.apply_event(&event_getter, &event).await?;
 
-            let attestation = self.node.policy_agent.check_event(&self.node, &self.cdata, &entity_before, &forked, &event)?;
+            let attestation = self.node.policy_agent.check_event(&self.node, &cdata, &entity_before, &forked, &event)?;
             let attested = Attested::opt(event.clone(), attestation);
 
             attested_events.push(attested.clone());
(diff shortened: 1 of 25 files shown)


dnorman:
@coderabbitai pause

PR #437, opened 2026-08-04
session hardening: atomic attach, total walks, and bookkeeping that cannot leak

dnorman:
Hardening for the session substrate that merged in #434, consolidating the confirmed findings of two independent post-merge reviews plus the automated review. Structure only; the reactions layer still arrives with #426.

**The attach cycle race.** `attach` checked reachability with no lock and `reaches` had no visited set, so two threads attaching opposite directions could both see "no cycle" and both install, after which the forwarded change signals loop without end. Now a module-level topology lock spans the reachability check and the install (attach is rare, so one lock across all credential types), `reaches` visits each set once so the walk is total on any graph, and the doc comments state the invariant actually established. The regression test was mutation-verified: with the lock removed, the racing test reproduces the stack overflow.

**The subscribe error path takes back what it minted.** A failed SubscribeQuery used to leave the freshly minted per-query session in the handler's map forever: the map insert preceded two fallible awaits, and `remove_predicate`'s early `?` on PredicateNotFound skipped the map cleanup. Now `remove_predicate` removes the entry unconditionally before propagating the reactor's result, and `subscribe_query` removes a vacant-minted entry on any error exit. One structural note: funneling nine `?` sites to a single cleanup exit meant wrapping the fallible body in an async block, so that file's diff is mostly re-indentation. One narrow race is known and deliberate: two in-flight subscribes for the same query id where the first fails can drop the map handle the second still uses; the second keeps a working clone and the next re-subscribe re-mints. Its real fix belongs to the claw-back work in #426.

**The registry stops being a public credential enumeration.** `Node.sessions` is now `pub(crate)`; tests observe it through a `test-helpers`-gated `session_registry()` accessor, following the existing pattern.

**Smaller items.** `Session::update` keeps its lock-free shape (a lock across the store would hold it across the broadcast, and subscribers may re-enter update); its comment now states the actual guarantee, verified against the signals crate: subscribers re-read at fire time, so a racing pair of identical updates costs at most one duplicate notification. The reactor hands out a subscription by id instead of the two-private-hop reach-through. The single-caller `cleanup_removed_query` helper with the false "shared" doc is inlined back. The server's refresh loop carries a debug assertion that the per-query set is single-member, which is true by construction. The wire-arm comment no longer describes this commit's own feature as future. The PolicyAgent docs now define the empty-credential-collection case plainly, including the honest statement that PermissiveAgent is permissive by design and where JwtAgent's fail-closed behavior actually comes from. session.rs's tests move to session/tests.rs per the new convention, and `/CLAUDE.md` joins .gitignore.

**Verification.** Full workspace battery: 812 passed, 0 failed across 124 binaries (baseline 810 plus the two new attach regressions). Review provenance: the trail files ankurah-434-postmerge-review-opus.md and ankurah-434-postmerge-review-codex.md carry the findings this PR discharges; the registry-superset and multi-session typed-read findings ride the upcoming authority-model PR instead, by design.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@
 .DS_Store
 .vscode
 *.code-workspace
+/CLAUDE.md
--- a/core/src/node.rs
+++ b/core/src/node.rs
@@ -141,7 +141,7 @@ where PA: PolicyAgent
     /// this node: each context attaches its credential source here at
     /// construction, so any state the node acts under is enumerable
     /// from one place.
-    pub sessions: SessionSet<PA::ContextData>,
+    pub(crate) sessions: SessionSet<PA::ContextData>,
 
     /// The reactor for handling subscriptions
     pub(crate) reactor: Reactor,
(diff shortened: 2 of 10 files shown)


dnorman:
Review round complete: independent Codex and Opus passes both confirmed one HIGH finding and cleared all the concurrency work (no lock inversion, no Drop deadlock; the racing regression was mutation-verified by both sides independently). The finding: the subscribe error funnel removed only the credential map entry, while a failure after the reactor upsert left the query installed with its gap fetcher holding the failed attempt's session - a later re-subscribe re-mints, register_or_get_query keeps the old fetcher, and gap fetches authorize under a frozen, untracked credential forever. Reachable single threaded (expand_states or delta generation failing) and via the client's selection-update re-subscribe path; a regression against main, where the map and fetcher could never disagree. Fixed in e6fb4b06: the error path now tears down both the map entry and the reactor query, guarded by set identity (ptr_eq) so an interleaved re-mint is never clobbered; the residual same-id concurrent case is documented in place and its generation guard belongs to the claw-back work in #426. The confirmed doc corrections rode along (funnel and remove_predicate comments, the topology lock's Drop-reentrancy hazard note for #426, one wording fix in the policy docs), and the test-helpers registry accessor now returns credential values, not writable session handles. Dismissed per maintainer ruling: the .gitignore CLAUDE.md entry stays (explicitly requested). Verified: core lib 219 passed, session suite 8 passed, session_cdata green, all-targets check clean with bench-internals,test-helpers.

PR #442, opened 2026-08-04
jwt-auth: events answer to the same row scopes as state

### LINKED ISSUE #438: jwt-auth: check_read_event authorizes at collection level only — per-entity read scopes are not evaluated for event reads
**Where:** `extensions/jwt-auth/src/agent.rs` — `check_read_event` calls `can_access_collection` and returns; the entity-state path (`check_read`) additionally builds a `TemporaryEntity` and runs `enforce_read_scope`. Verified against the 0.9.0 source.

**Consequence:** for any collection carrying a row-level read scope, entity *state* is correctly filtered per-row, but that entity's *events* are readable by any token that passes the collection-level gate — `GetEvents` becomes a path around the row scope. Since events for CRDT-backed properties reconstruct content, the row scope is not protective for such collections.

**Downstream context (ankurah/community):** community currently has four read-scoped collections live (`ban`, `readstate`, `notification`, `notificationpref`) — the mismatch was consciously accepted for that data's sensitivity under the earlier reachability posture. Two-party DMs (ankurah/community#30) are explicitly **blocked on this fix**: dm_message content is exactly what makes the event path unacceptable. Community will pin the release containing the fix before shipping any dm_* collection.

**Acceptance:** an adversarial test in the jwt-auth suite proving a token that fails an entity's read scope is denied that entity's **events** — and ideally coverage across all three read paths (state fetch, event fetch, live subscription delivery), since the reactor path shares the enforcement seam.

**Sizing note:** the fix looks contained — `check_read_event` has the event's collection and id in hand; evaluating the same scope rules the state path uses (against current state, or denying when state is unavailable) mirrors the existing `check_read` structure. Happy to take this in-lane if wanted; flagging rather than presuming.

dnorman:
Fixes #438: `check_read_event` authorized at collection level only, so for any collection carrying a row-level read scope, entity STATE was filtered per row while that entity's EVENTS were readable by any token passing the collection gate — and events reconstruct CRDT content, so GetEvents was a path around the row scope.

**The fix.** One private `check_read_entity` holds the whole row-level read decision (collection gate, privileged bypass, no-scope-rules early out, TemporaryEntity, `enforce_read_scope`), and both `check_read` and `check_read_event` delegate into it — the event path inherits the enforcement's semantics (including #436's ruled skip behavior) rather than restating them. When a scoped collection has no current state to evaluate, the read refuses; unscoped collections keep their existing collection-gate behavior. Both halves pinned by tests.

**Three guards make the guarantee hold under adversarial review** (two independent review passes, findings adjudicated and applied):

- **The pairing is a contract, not a convention.** `check_read_event` takes `Option<&proto::EntityState>` and the agent refuses a state whose entity id differs from the event's. The trait doc is normative: the caller MUST supply the state of the event's own entity.
- **A state only authorizes events it can speak for.** Commits publish events before persisting state, so a served event could postdate the state that authorized it — including the very event that moves the row out of the caller's scope. GetEvents now serves an event only if the authorizing state's head causally covers it (head membership is O(1); anything older costs a DAG walk proportional to the distance, budget-capped and fail-closed on ambiguity — a peer backfilling deep history one event per request pays that depth per call, noted in the method doc). The node's resident entity outranks the stored state buffer, which closes the indefinite variant of the race under the EventOnly apply path.
- **The event bridge walks one entity's history or none.** A graft in a readable entity's ancestry can reference another entity's events; the walk now excludes any fetched event whose entity id differs from the bridge's entity and falls back to the existing state-snapshot path, so a foreign event is never checked against the wrong row's state.

**Signature ripple, contained:** the trait change touches PermissiveAgent, core's two call sites, and four test-only agents. GetEvents batches states over the distinct entities in the served set (the default `get_states` walks one entity at a time — K distinct entities cost K sequential reads until an engine grows a real multi-key read).

**Accepted residuals, documented in the spec and trait rather than implied:** event reads are authorized against the CURRENT row state, so a row's history from before a scope change remains readable to a caller who can read the row today — per-event historical evaluation is tracked in #445. Separately, the IndexedDB engine stores events without collection identity, so on that engine an event can be requested through the wrong collection past every gate — engine fix tracked in #444. The live subscription delivery path (the reactor's StateAndEvent) rides the #426 recomposition, per the plan.

**Tests.** The original wire-path pair (a token's GetEvents returns exactly the allow/deny split its Get returns, asserted in both directions; missing state refuses on scoped collections, spares unscoped ones) plus: privileged caller, plural-credential union on the event path, an event the authorizing state does not cover is refused, a resident entity outranks a rewound stored state, and a forged cross-entity ancestry does not bridge (via a test-helpers-gated bridge wrapper, disclosed). Every guard was mutation-verified — reverting each makes its test fail on the leak assertion. Full workspace battery: 819 passed, 0 failed.

### DIFF
--- a/core/src/lazy_entity_getter.rs
+++ b/core/src/lazy_entity_getter.rs
@@ -1,0 +1,60 @@
+//! Lazy, single-entity lookup for policy checks: a check receives a getter
+//! bound to one entity and fetches the row only if its verdict turns on row
+//! content, so verdicts that never look at the row never pay a read.
+
+use crate::{
+    entity::Entity,
+    error::RetrievalError,
+    node::Node,
+    policy::PolicyAgent,
+    retrieval::{LocalEventGetter, LocalStateGetter},
+    storage::StorageEngine,
+};
+use ankurah_proto::{CollectionId, EntityId};
+
+/// Fetches one preordained entity, on demand. Which entity is fixed at
+/// construction; [`Self::get`] takes nothing and fetches it, so a check that
+/// is handed one of these cannot be pointed at any other row, and a check
+/// that never calls `get` costs nothing. Construct via
+/// [`Node::entity_getter`](crate::node::Node::entity_getter).
+pub struct EntityGetter<'a, SE, PA>
+where PA: PolicyAgent
+{
+    entity_id: EntityId,
+    collection: CollectionId,
+    node: &'a Node<SE, PA>,
+}
+
+impl<'a, SE, PA> EntityGetter<'a, SE, PA>
+where
+    SE: StorageEngine + Send + Sync + 'static,
+    PA: PolicyAgent + Send + Sync + 'static,
+{
+    pub(crate) fn new(node: &'a Node<SE, PA>, entity_id: EntityId, collection: CollectionId) -> Self {
+        Self { entity_id, collection, node }
+    }
+
+    /// The entity this getter is bound to.
+    pub fn id(&self) -> EntityId { self.entity_id }
+
+    /// The entity as it currently stands, or `None` when nothing current
+    /// exists for it. The resident entity when the node holds one, otherwise
+    /// materialized from stored state into the node's entity set, which is
+    /// what makes repeated calls not repeat the lookup.
+    pub async fn get(&self) -> Result<Option<Entity>, RetrievalError> {
+        let collection = self.node.collections.get(&self.collection).await?;
+        let state_getter = LocalStateGetter::new(collection.clone());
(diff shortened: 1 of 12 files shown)


dnorman:
Parked, potentially sidelined by a larger initiative: the event security model is being redesigned comprehensively (per-event visibility vs current-row verdicts, event structure changes incl. attested author and genesis entropy, and PR #359's decomposition) targeting 0.10. This branch holds the WIP state of the lazy-getter approach; its staged causal-coverage test is deliberately red pending that design. Nothing depends on this landing now; community#30 remains blocked on the eventual fix rather than this PR.

PR #457, opened 2026-08-07
planner: bounds are a prefilter; the residual re-checks every term

dnorman:
## The defect

For a conjunction with two terms on one column — `owner = 'bob' AND owner = 'alice'` — the planner built an index key of `[owner, owner]`, bound **both** positions to the first term's value (`['bob','bob']`), and dropped **both** conjuncts from the residual predicate, because `calculate_remaining_predicate` consumed conjuncts by column name alone. The second term was neither encoded into the range nor left to re-check, so the scan answered with the first term's rows: a query whose correct answer is empty returned rows it excluded. The same by-column-name consumption lost a caller's range term when an equality on the same column took the key position (`owner >= 'bob' AND owner = 'alice'` scanned `alice..alice` and forgot the range).

## Why "the bounds enforce that term" is not a thing the planner can claim

The first commit on this branch fixed the shape above with a receipt: bounds returned the exact `(column, operator, value)` comparisons they encode, and only receipted conjuncts were elided. Review found the receipt still dishonest, one layer down, and the second commit follows that finding to its end.

The planner types each index keypart **from the query literal** — `TypeResolver::resolve_path` has no schema to consult for a simple path — and the key encoding (`encode_component_typed`) casts every **stored value** to that type when building index keys. A cast can coalesce values the predicate distinguishes: integer literals parse as `I32`, `F64 → I32` truncates, so a stored `score` of `5.9` keys as `5` and sits **inside** the point range for `score = 5`, while the predicate evaluator — which casts the literal toward the stored value's type instead — rejects it. The bounds enforce "encodes like the literal"; the predicate means "compares equal to the literal". Those differ whenever the stored type differs from the literal's, which is an everyday event (`score = 5` against an f64 column), and the planner cannot see stored types from where it sits. `score = 5 AND score = 5` returned the 5.9 row with nothing left to stop it — and so did `score = 5` alone.

## The contract now

**Bounds are a prefilter; the residual is the truth.** Every index plan carries the whole predicate as its residual — exactly what the TableScan fallback always did — and both engines already re-check it per row (sled `exec_index_scan_plan`, indexeddb `execute_plan_query`). Bounds still narrow the scan: point ranges from the first equality per column, folded low/high endpoints for ranges, `push_keypart_once` keeping a column to one key position. No conjunct is ever elided. When schema-derived keypart typing exists, elision can return with a proof; today no such proof is available to the planner, so it does not claim one.

## Shapes covered

| shape | outcome |
|---|---|
| two equalities, one column (`owner='bob' AND owner='alice'`) | fixed — bounds `bob..bob`, residual carries both terms, answer empty |
| identical duplicate (`owner='bob' AND owner='bob'`) | correct rows, one key position; residual retained (see cost note) |
| equality + range, one column (`age=5 AND age>10`, both orders) | fixed — point bound, residual carries both, answer empty |
| two ranges, one column (`age>3 AND age<10`) | bounds fold both endpoints; residual still re-checks |
| cross-typed ranges (`age>5 AND age>'abc'`) | bounds narrow by the first comparable term; residual carries both |
| integer literal vs f64 column (`score=5` and `score=5 AND score=5`, stored `5.9`) | fixed — key range admits the truncated key, residual rejects the row |
| `!=`, `IN`, `BETWEEN`, disjunctions | never bounds material; enforced by the residual as before |
| `Ref`-typed and `String`-typed columns | both covered at sled and jwt-auth layers |

## Consequence worth naming

jwt-auth composes a policy read scope onto the caller's predicate as another conjunct, so a scope of `owner = $jwt.sub` with a caller filtering `owner = <someone else>` is exactly the two-equalities shape — the scope's term was the dropped one, and a fetch returned rows outside the caller's read scope. `extensions/jwt-auth/tests/scope_column_overlap_tests.rs` pins that composition end to end (fetch and LiveQuery, `Ref` and `String` scoped columns, disjunctive two-arm scopes).

## Pins updated deliberately (they pinned the elision)

- `tests/tests/sled/collection_boundary.rs::test_prefix_guard_toggle_effect` pinned the scan **overshoot appearing in results** when the debug prefix guard is off — observable only while the planner dropped the equality residual. It now pins the answer staying right with the guard off; the guard remains a scan-narrowing optimization.
- The json-path "full pushdown" assertions (`storage/common/src/planner.rs`, `tests/tests/sled/json_property.rs`) pinned `remaining_predicate == True`; they now pin the residual re-checking the term.
- `every_plan_carries_the_whole_predicate_as_residual` asserts the **exact** residual for every plan under both configs; its predecessor accepted any non-`True` residual, which a regressed plan dropping the decisive term (residual `age = 5` without `age > 10`) would still have satisfied.

## Cost accepted

The `remaining_predicate == True` no-filter path in sled now fires only for queries with no comparison terms (e.g. unfiltered ORDER BY scans). Every indexed fetch pays a per-row materialized-values lookup plus one predicate walk over the rows the bounds admit — the work the TableScan path always did, on a bounds-narrowed row set. That is the price of answers that do not depend on the literal's type happening to match the stored value's type.

## Not addressed here (pre-existing, adjacent)

- A scan can still **miss** rows whose stored type differs from the keypart type (`F64(5.9)` under an `I32` keypart keys as `5`, so `score > 5` skips it; a `String` row under a numeric keypart fails the cast and is absent from the index). False negatives cannot be repaired by a residual; they need schema-typed or type-tagged index keys.
- `EmptyScan` still trusts a decisively contradictory same-typed fold (`x > 10 AND x < 5`), which a cross-typed stored value could satisfy under the evaluator's casts (`String("3")` compares `> "10"` and `< "5"` lexicographically).

Both predate this branch and are the remaining cast-related gaps once no term can be silently dropped.

## Evidence

- `tests/tests/sled/repeated_column.rs::f64_column_with_integer_literal_matches_exactly` fails on the first commit and on `main` (`score = 5` returns `["alice note"]`, the stored 5.9 row) and passes here; its controls pin that `score = 2` still matches a stored `2.0` (no over-filtering).
- `every_plan_carries_the_whole_predicate_as_residual` fails against the first commit's planner (partial residual `owner = 'alice'` where the whole predicate is demanded) and against `main` (residual `True`).
- The original defect-shape tests (planner unit, sled rows, jwt-auth composition) all fail on `main` as before: e.g. `two_equalities_on_one_column` observes keyparts `[__collection, owner, owner]`, both bounds `bob..bob`, `remaining_predicate: True`; sled `two_equalities_on_one_ref_column_match_nothing` gets `["bob note"]` where `[]` is demanded; `caller_equality_on_the_scoped_column_does_not_replace_the_scope` returns the other member's row.

Gates, all green: `cargo fmt --all -- --check`; `cargo test -p ankurah-core -p ankurah-jwt-auth` (the pins from #456 hold — their evaluator-semantics pins are what make the residual well-defined for cross-typed comparisons); `cargo test -p ankurah-storage-common -p ankurah-storage-sled -p ankurah-tests` (60/60 suites); `cargo check --tests`.

## Engines

The change is in the shared planner used by sled (`PlannerConfig::full_support`) and indexeddb (`PlannerConfig::indexeddb`); both engines already apply `remaining_predicate` per row. Postgres and sqlite never call this planner — they split predicates into SQL plus a residual — and are untouched.

### DIFF
--- a/extensions/jwt-auth/tests/scope_column_overlap_tests.rs
+++ b/extensions/jwt-auth/tests/scope_column_overlap_tests.rs
@@ -1,0 +1,254 @@
+//! What a read scope does when the caller's own filter names the same column.
+//!
+//! A read scope is composed onto the caller's predicate as another conjunct, so
+//! `{ "filter": "owner = $jwt.sub" }` and a caller asking for `owner = <someone
+//! else>` produce one predicate with two terms on `owner`. Nothing about that
+//! is special to policy — it is the ordinary case of a column carrying more
+//! terms than an index key has positions for it, pinned in the storage planner
+//! by `repeated_column_tests` in `storage/common/src/planner.rs` and over sled
+//! rows by `tests/tests/sled/repeated_column.rs`.
+//!
+//! It gets its own coverage here because this is the composition consumers
+//! actually run, and because it is the case where a term the planner forgets is
+//! visible as a row: an index scan that answers a two-term query using only the
+//! first term returns rows the second term excludes. The checks run against a
+//! sled-backed durable node over both paths a reader takes — a one-shot `fetch`
+//! and a LiveQuery.
+
+mod common;
+
+use ankurah::{Model, Node, Ref};
+use ankurah_jwt_auth::{JwtAgent, JwtContext, JwtKeys, PolicyConfig, SigningKeys};
+use ankurah_storage_sled::SledStorageEngine;
+use std::sync::Arc;
+
+#[derive(Model, Debug, serde::Serialize, serde::Deserialize)]
+pub struct OverlapUser {
+    pub name: String,
+}
+
+/// A scoped column that is a `Ref`, which collates as raw EntityId bytes.
+#[derive(Model, Debug, serde::Serialize, serde::Deserialize)]
+pub struct OverlapDoc {
+    pub owner: Ref<OverlapUser>,
+    pub label: String,
+}
+
+/// The same scope over a String column, which collates as text and admits range
+/// comparisons a `Ref` column does not usefully take.
+#[derive(Model, Debug, serde::Serialize, serde::Deserialize)]
+pub struct OverlapNote {
+    pub owner: String,
+    pub label: String,
+}
+
+/// Two scoped columns under one disjunctive rule — the shape a direct-message
+/// thread takes, where either participant may read.
+#[derive(Model, Debug, serde::Serialize, serde::Deserialize)]
+pub struct OverlapThread {
+    pub first: Ref<OverlapUser>,
+    pub second: Ref<OverlapUser>,
+    pub label: String,
+}
+
+const CONFIG_JSON: &str = r#"{
+    "roles": { "Member": ["doc:read", "user:read"] },
+    "collections": {
+        "overlapdoc":    { "read": "doc:read", "scope": [{ "filter": "owner = $jwt.sub" }] },
+        "overlapnote":   { "read": "doc:read", "scope": [{ "filter": "owner = $jwt.sub" }] },
+        "overlapthread": { "read": "doc:read", "scope": [{ "filter": "first = $jwt.sub OR second = $jwt.sub" }] },
+        "overlapuser":   { "read": "user:read" }
+    }
+}"#;
+
+fn member_context(keys: &SigningKeys, subject: &str) -> JwtContext {
+    let claims = common::make_claims(subject, &["Member"], "member@example.com");
+    let token = common::sign_token(keys, &claims);
+    JwtContext::from_claims(claims, token)
+}
+
+/// Alice, Bob and Carol; one document, note and thread per pairing, on a node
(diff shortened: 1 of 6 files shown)


dnorman:
A field instance from Ankurah Community, while shaping a moderator-only read scope on sled (0.9.2), consistent with this issue's term-drop family and adding one datum: the leak reproduces with two terms on ONE property.

Scope rule `resolved = false AND resolved = true` (deliberately unsatisfiable, as a probe): the live/subscription path delivered nothing and the by-id read was refused — both correct — while the in-process durable `fetch` returned the whole collection (`fetch rows = 1` against a 1-row collection, by-id denied). Both terms are plain equalities on the same column, so this looks like the first-match-per-keypart / consumed-by-column-name-only shape rather than a residual-recheck gap.

Operational lesson now pinned in that repo's policy tests: a scope rule must not constrain the same property twice until this lands. No action needed from this side — noting it here so the fix's test matrix can include the two-terms-one-property case.

PR #460, opened 2026-08-07
jwt-auth, core: a read names its kind, and `get` admits by-id reads alone

dnorman:
## Why

Community wants user rows public to whoever can name them while the roster stays signed-in-only: message authors carry user ids, so a guest rendering display names needs by-id loads and nothing else. The wire already tells the shapes apart (`Get{ids}` vs `Fetch`/`SubscribeQuery{selection}`); the policy layer could not — one collection gate admitted every read.

## What

`ReadKind::{Get, Scan}` travels with every collection gate and per-row read check. **Get** is entity state addressed by an id the caller presented; **Scan** is anything whose answer names rows for the caller — predicate matches, subscription deliveries, event backfill. `GetEvents` and `check_read_event` rule themselves Scan on those terms: an event log carries history beyond any one presented id.

The jwt-auth config grows an optional per-collection `get` privilege. `read` or `write` admit every kind, as before; `get` additionally admits Get alone; a policy without the field reads the same as before the field existed — by-id reads keep requiring `read`. In `filter_predicate` and `enforce_read_scope`, a credential counts toward an answer only at the kind it is admitted for: a get-tier credential vets by-id rows and never widens a query it could not run.

## Coverage

`extensions/jwt-auth/tests/get_privilege_tests.rs`: the split (view reaches a user row by id and never by scan), the fallback (absent `get`, no weakening), read admitting both kinds, and end to end through a node — a guest loads the row it names, is refused both `fetch("true")` and `fetch("name = 'Ada'")` (a narrowing predicate is still a scan), and the member keeps the roster.

Gates: fmt check clean; core 220, jwt-auth all suites (incl. the 4 new), storage-common 106, sled, and ankurah-tests green locally (`test_websocket_subscription_propagation` tripped once under full-suite load and passed 3/3 isolated — the known websocket timing flake family; the change adds a parameter whose value is constant per call site, and every agent in that harness is permissive).

First consumer: community's `policy.json` sets `user` to `get: view, read: signed_in`, and ankurah-chat resolves author names per-ref instead of through a roster query.

### DIFF
--- a/core/src/context.rs
+++ b/core/src/context.rs
@@ -6,7 +6,7 @@ use crate::{
     livequery::{EntityLiveQuery, LiveQuery},
     model::View,
     node::{MatchArgs, Node},
-    policy::{AccessDenied, PolicyAgent},
+    policy::{AccessDenied, PolicyAgent, ReadKind},
     storage::{StorageCollectionWrapper, StorageEngine},
     transaction::Transaction,
 };
@@ -380,7 +380,7 @@ where
             debug!("Node({}).get_entity found local entity - returning", self.node.id);
             let state = local.to_state()?;
             let entity_id = local.id();
-            self.node.policy_agent.check_read(&cdata, &entity_id, collection_id, &state)?;
+            self.node.policy_agent.check_read(&cdata, &entity_id, collection_id, &state, ReadKind::Get)?;
             return Ok(local);
         }
         debug!("{}.get_entity fetching from storage", self.node);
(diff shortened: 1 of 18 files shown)


dnorman:
Superseded by the lookup-privilege design (enforcement in filter_predicate where the predicate is, zero core changes) — see the replacement PR referenced above. The ReadKind threading through core that this PR carried is retracted: with scans refused at predicate composition, a wire-shape enum in the trait has no remaining justification.

